Issue #10, opened 2018-08-20
Fix rectified linear neurons


tbekolay:
In several places we [try to make ReLus work](https://github.com/nengo/nengo-loihi/blob/master/nengo_loihi/loihi_cx.py#L122), but decoder solving does not work correctly, as shown in the example by @drasmuss:

```python
import matplotlib.pyplot as plt
import nengo
import numpy as np

import nengo_loihi

with nengo.Network() as net:
    a = nengo.Node(output=nengo.processes.WhiteSignal(1, 5, rms=0.3, seed=0),
                   size_out=3)
    ens = nengo.Ensemble(100, 3, neuron_type=nengo.RectifiedLinear(),
                         max_rates=nengo.dists.Uniform(100, 120),
                         intercepts=nengo.dists.Uniform(-0.95, 0.95))
    nengo.Connection(a, ens, synapse=None)
    out = nengo.Node(size_in=1)
    nengo.Connection(ens, out, synapse=None, function=lambda x: np.sum(x))
    p_a = nengo.Probe(a, synapse=0.1)
    p_out = nengo.Probe(out, synapse=0.1)

with nengo_loihi.Simulator(net, target="sim") as sim:
    sim.run(1.0)

plt.figure()
plt.plot(np.sum(sim.data[p_a], axis=-1), label="target", color="blue")
plt.plot(sim.data[p_out], label="output", color="orange")
plt.legend()
plt.show()
```

This might be magically fixed by using `nengo.SpikingRectifiedLinear`, but if not then we should track down and fix the issue.

drasmuss:
Fixed in #42

PR #24, opened 2018-08-21
Neuron to neuron examples

tbekolay:
A notebook with three example networks using direct neuron-to-neuron connections. Here are the plots from the notebook if you don't want to run it yourself:

* Simple fan-out with Nengo

1-nengo

* Simple fan-out with Nengo Loihi

1-loihi

* One-to-one connections with Nengo

2-nengo

* One-to-one connections with Nengo Loihi

2-loihi

* Fixed probability connections with Nengo

3-nengo

* Fixed probability connections with Nengo Loihi

3-loihi

This was originally going to be part of #23 (and therefore is based on it), but I ran into an issue so I figured I would split this off into a separate PR so that it doesn't block the release (if we manage to figure it out in time though, we can include it).

The issue is that when making neuron-to-neuron connection in Nengo Loihi, the transform matrix has to be scaled up by `1 / dt` (at least, scaling up by 1000 worked, so I assume it's a `1 / dt` scaling problem). This means that you can't run the same model in Nengo and expect it to work in Nengo Loihi if you're using direct neuron-to-neuron connections, which is a pretty big issue.

I'll start looking into this tomorrow, but if anyone knows why this might be the case (or how it can be fixed), help would be appreciated.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,5 +36,6 @@ docs/examples/integrator.py
 docs/examples/integrator_multi_d.py
 docs/examples/keyword_spotting.py
 docs/examples/learn_communication_channel.py
+docs/examples/neuron_to_neuron.py
 docs/examples/oscillator.py
 docs/examples/oscillator_nonlinear.py
--- a/docs/examples.rst
+++ b/docs/examples.rst
@@ -15,5 +15,6 @@ We compare performance with Nengo core where appropriate.
    examples/integrator_multi_d
    examples/oscillator
    examples/oscillator_nonlinear
+   examples/neuron_to_neuron
    examples/learn_communication_channel
    examples/keyword_spotting
--- a/docs/examples/neuron_to_neuron.ipynb
+++ b/docs/examples/neuron_to_neuron.ipynb
@@ -1,0 +1,333 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Neuron to neuron connections\n",
+    "\n",
+    "While Nengo is often used with deep learning\n",
+    "and NEF style networks,\n",
+    "it can also be used for lower level models\n",
+    "in which each neuron to neuron connection\n",
+    "is explicitly specified.\n",
+    "\n",
+    "In these examples, we connect a `pre` population\n",
+    "to a `post` population with different sets of\n",
+    "specified connection weights."
+   ]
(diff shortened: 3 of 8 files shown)


drasmuss:
There used to be a 1/dt scaling on neuron connections, but I had to remove
it to get the other neuron-neuron test to work properly (the amplitude
one). I notice that one difference is that that test has synapse=none and
these ones don't, so maybe it has something to do with how synapses are
implemented? E.g. perhaps the dt scaling is being implemented in the
synaptic filtering, rather than the voltage accumulation, so you get a
dramatic change in behaviour with synapse=none? That's just a guess.

drasmuss:
Update: figured out it has to do with the `RectifiedLinear` implementation.  Currently `RectifiedLinear` does not scale its voltage accumulation by `dt` (which is why we needed to scale the connection weights by `dt` to get the correct behaviour).  This is probably related to #10.  Going to see if I can figure out how to fix `RectifiedLinear`, but in case it isn't straightforward I'd say we just put the `1/dt` scaling on connection weights back in and `xfail` the RectifiedLinear tests, so that we can get this merged.

tbekolay:
Agreed, better to release without ReLu support than with this bug.

tbekolay:
OK I pushed a commit disabling ReLu and the associated tests, and reverting the weight scaling change commit. With this, the neuron-to-neuron connections look good in the emulator; running things on Loihi now.

tbekolay:
This also I would like in the release I'm planning for Monday so could someone please review this soon.

drasmuss:
I was trying to figure out why the `test_ens_neurons_on_host` test was failing, but no success yet.  I think it's OK not to block the release on that, but maybe we could put the old test back in (in place of or in addition to the reverted one) and xfail it so we have a note/test case showing something weird going on there?

tbekolay:
Yeah sounds good, I'll add in the old one alongside this one.

tbekolay:
After a bit of discussion, @drasmuss and I are pretty sure that the reason the old test is failing is because seeds change when moving the ensemble on/off chip, which is the main difference between the two tests (one sampled the parameters beforehand, the other let Nengo sample the parameters). Instead of adding back the old test, which is mostly a duplication of this one, I added a parameter as to whether we should seed the ensemble itself and `xfail`ed the case when the ensembles weren't seeded but the network is seeded.

PR #25, opened 2018-08-23
Add INRC-specific docs

tbekolay:
It seems as though most people will be interacting with Loihi through the INRC superhost. Most of the docs are fine if you substitute INRC when we say "superhost" (it's likely a good thing for people to learn the term superhost well), but the installation instructions need to be different for a few reasons. Making a separate INRC section made more sense to me than making the generic superhost instructions have two possible paths. Also added INRC to the overview terminology and made some slight modifications in a few places. Also also added the IPython / Jupyter config thing that we had to do on the INRC to get it working.

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -25,6 +25,127 @@ Follow `our NumPy install instructions
 <https://www.nengo.ai/nengo/getting_started.html#installing-numpy>`_,
 then try again.
 
+INRC
+====
+
+.. note:: These instructions will only work on the INRC superhost.
+          If you have set up your own superhost
+          as per the :doc:`setup/superhost` page,
+          then see the next section.
+
+These steps will take you through
+setting up a Python environment
+for running Nengo Loihi,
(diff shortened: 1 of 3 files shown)


tbekolay:
Also note this is based on #24, as we might as well have both in the release given that we're still waiting on the release announcement OK.

tbekolay:
Hm... as it turns out, running the snips with `conda` might indeed be possible, it might be an issue with the `ncl-ext-mh-01` board ... meaning a big chunk of this PR can be omitted ... sigh. Will update when I hear back from Intel...

tbekolay:
They haven't looked into it yet so I think to be safe we just go with this PR. We can always simplify the docs later; the virtual env installation will definitely work so might as well recommend people do that.

Also the release announcement is good to go, so if someone could review ASAP it would be appreciated, my plan is to release Monday morning so if no one reviews I'm just going to merge unreviewed, typos and all

PR #27, opened 2018-08-24
Better stopping behaviour for snips

tcstewar:
We now send an explicit stop signal, and once that is received the snip will just return immediately.

We needed better stopping behaviour for working with nengo_gui (since we don't know when the simulator might be closed).  It also handles pressing Ctrl-C while your model is running, which might be handy.

[EDIT:]  This PR may need to be re-thought in terms of making it fit with #26 , which made some big changes to the communication protocol between chip and host.

### DIFF
--- a/nengo_loihi/simulator.py
+++ b/nengo_loihi/simulator.py
@@ -319,7 +319,6 @@ class Simulator(object):
 
     def step(self):
         """Advance the simulator by 1 step (``dt`` seconds)."""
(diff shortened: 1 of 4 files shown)


tbekolay:
Since #28 includes the main commit in this branch, I'm going to close this and deal with reviewing that commit in #28. The extra ones I added here I will move into a separate PR.

PR #30, opened 2018-08-27
Rearrange main classes, refactor emulator

tbekolay:
This does the code moves and renames that we had previously discussed and refactors the emulator.

Still a few things todo:

- [ ] The emulator / probe signal renames likely broke the hardware backend but due to some environment issues it was hard to test last week.
- [ ] Currently this is based pretty far back in the master history, so I'll need to rebase.


nengo_loihi/builder.py
```diff
@@ -731,6 +524,7 @@ def signal_probe(model, key, probe):
     model.objs[probe]['out'] = cx_probe
 
 
+# TODO: why q, s, v, u ?
```
hunse: These correspond to the variable names used in `CxSimulator`, which in turn were based off those in the mathematical overview of the chip logic provided by Intel.

nengo_loihi/builder.py
```diff
@@ -21,8 +21,14 @@
 from nengo.utils.compat import iteritems
 import nengo.utils.numpy as npext
 
+from nengo_loihi.group import CoreGroup
```
hunse: To me, `CoreGroup` sounds like a group of cores (which it's not). What about `NeuronGroup`, since it pretty much represents a group of neurons?

nengo_loihi/builder.py
```diff
@@ -27,9 +27,9 @@
     CxModel,
     CxProbe,
     CxSpikeInput,
-    CxSynapses,
 )
 from nengo_loihi.splitter import ChipReceiveNeurons, ChipReceiveNode
+from nengo_loihi.synapses import CxSynapses
```
hunse: I'm wondering if we should have a `model` folder, which will hold all the formerly `Cx` object files, such as this one. Just to make it clear that they all go together to describe this level of model.

nengo_loihi/builder.py
```diff
@@ -29,7 +29,7 @@
     CxSpikeInput,
 )
 from nengo_loihi.splitter import ChipReceiveNeurons, ChipReceiveNode
-from nengo_loihi.synapses import CxSynapses
+from nengo_loihi.synapses import Synapses
```
hunse: So one nice thing about having the `Cx` on the front is that it made it clear that this `Synapses` object was different from `nengo.Synapse`.

If we have that `model` folder I was talking about, then here I would just `import nengo_loihi.model as loihi_model`, so that we can be more explicit with `loihi_model.Synapses`, etc.

tbekolay:
Because this touches a lot of things that are also touched in other open PRs (and since this is based on an earlier point in mater anyhow) we're going to put this on hold for now and redo it once the current PR queue is cleared.

tbekolay:
As mentioned in the last standup, this PR's back on hold until after the next release.

drasmuss:
This was superseded by #159

Issue #31, opened 2018-08-27
Check that Python >= 3.5


hunse:
In nengo_loihi's `__init__` file, I think we should check that the python version is new enough. Not only do we need this for working with NxSDK, but we also have a star expression in `simulator.py` (`items.append((sent_count, *output_axon[j]))`) that uses new Python syntax. Right now, if you try to import with older Python, you get a somewhat-cryptic syntax error.

tbekolay:
Right now NxSDK needs < 3.6 too, so if we're going to limit it based on NxSDK we should check for both. The emulator doesn't have those restrictions though, so should we fix the Python 2 compat stuff instead? There probably isn't much to it and if people are just trying things out might be worth having it work in Python 2. But I'm fine either way, also convenient to not have to deal with Py2/3 in a single codebase.

hunse:
So the emulator now works with all of Python 3.

Do we want any checks for `nxsdk`? Really, this is something that `nxsdk` should be checking in it's `setup.py` file, but as far as I know it still doesn't have a `setup.py` file. I'm tempted to say "no" to this, and close this issue for now.

tbekolay:
Yeah I would leave that up to `nxsdk` and consider this issue closed.

PR #37, opened 2018-08-30
Compatibility with other python versions

hunse:
Addressing #31.

Removing the star expression allows the emulator to work with Python 3.4. The tests also all appear to pass on Python 2.7, though I did not add any `__future__` import statements, so it's not guaranteed e.g. division will work identically in Python 2.

I mostly use Python 3 now, so I'm happy to just merge this and only officially support Python 3 (but ideally all common versions of it) going forward. However, if you think Python 2 support is also good to have, speak up!

Also, I'm tempted to say we should either officially support Python 2, or check on import that we're using Python 3.

### DIFF
--- a/nengo_loihi/simulator.py
+++ b/nengo_loihi/simulator.py
@@ -481,8 +481,8 @@ class Simulator(object):
                         for j, s in enumerate(spike_input.spikes[sent_count]):
                             if s:
                                 for output_axon in spike_input.axon_ids:
-                                    items.append((
-                                        sent_count, *output_axon[j]))
+                                    items.append(
+                                        (sent_count,) + output_axon[j])
                         sent_count += 1
                     spike_input.sent_count = sent_count
                 if len(items) > 0:
--- a/setup.py
+++ b/setup.py
@@ -34,6 +34,7 @@ setup(
     description="Run Nengo models on the Loihi chip",
     long_description=read("README.rst"),
     zip_safe=False,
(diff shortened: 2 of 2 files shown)


nengo_loihi/simulator.py
```diff
@@ -481,8 +481,8 @@ def handle_host2chip_communications(self):  # noqa: C901
                         for j, s in enumerate(spike_input.spikes[sent_count]):
                             if s:
                                 for output_axon in spike_input.axon_ids:
-                                    items.append((
-                                        sent_count, *output_axon[j]))
+                                    items.append(
+                                        (sent_count,) + output_axon[j])
```
tbekolay: Is `output_axon[j]` a tuple? Does it get automatically converted to a tuple because the left hand side is a tuple? My instinct is to ensure the right hand side is also a tuple, but if Python does that automatically then it's no big deal.

drasmuss:
Python 3 only sounds good to me.

nengo_loihi/simulator.py
hunse: So Python will not automatically convert (adding a list and a tuple results in an error, when done in either order, at least in Python 3). I'm pretty sure that `output_axon` is a tuple. (I recall testing this and having it run on Loihi, but somebody else should double-check.)

hunse:
I'm happy with Python 3 only as well. Having the check in `setup.py` is fine by me.

Issue #32, opened 2018-08-27
Better error message if no objects on chip


tbekolay:
Just ran into this error when I marked the only ensemble in a network as one to simulate off chip:

```
self = <nengo_loihi.loihi_cx.CxSimulator object at 0x7f1898e647f0>
model = <nengo_loihi.builder.Model object at 0x7f1898d8deb8>, seed = 434690618

    def build(self, model, seed=None):  # noqa: C901
        """Set up NumPy arrays to emulate chip memory and I/O."""
        if seed is None:
            seed = np.random.randint(2**31 - 1)
    
        logger.debug("CxSimulator seed: %d", seed)
        self.seed = seed
        self.rng = np.random.RandomState(seed)
    
        self.t = 0
    
        self.model = model
        self.inputs = list(self.model.cx_inputs)
        self.groups = sorted(self.model.cx_groups,
                             key=lambda g: g.location == 'cpu')
        self.probe_outputs = collections.defaultdict(list)
    
        self.n_cx = sum(group.n for group in self.groups)
        self.group_cxs = {}
        cx_slice = None
        i0 = 0
        for group in self.groups:
            if group.location == 'cpu' and cx_slice is None:
                cx_slice = slice(0, i0)
    
            i1 = i0 + group.n
            self.group_cxs[group] = slice(i0, i1)
            i0 = i1
    
        self.cx_slice = slice(0, i0) if cx_slice is None else cx_slice
>       self.cpu_slice = slice(self.cx_slice.stop, i1)
E       UnboundLocalError: local variable 'i1' referenced before assignment
```

`i1` is unbound because there are no groups, i.e., nothing being simulated on Nengo Loihi. I think it probably makes sense to raise a runtime error if we're about to simulate something completely with Nengo and not using Loihi at all. The alternative would be to raise a warning and simulate anyway, but I think warnings are too easy to miss.

drasmuss:
Done in #38

PR #38, opened 2018-08-30
Check that some model part is run on-chip

hunse:
Addresses #32.

### DIFF
--- a/nengo_loihi/loihi_cx.py
+++ b/nengo_loihi/loihi_cx.py
@@ -5,7 +5,7 @@ import logging
 import warnings
 
 import numpy as np
-from nengo.exceptions import BuildError
+from nengo.exceptions import BuildError, SimulationError
 from nengo.utils.compat import is_iterable, range
 
 from nengo_loihi.loihi_api import (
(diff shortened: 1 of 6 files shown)


nengo_loihi/tests/test_simulator.py
```diff
@@ -0,0 +1,15 @@
...
+
+def test_cx_model_validate_notempty(Simulator):
+    with nengo.Network() as model:
+        nengo_loihi.add_params(model)
+
+        a = nengo.Ensemble(10, 1)
+        model.config[a].on_chip = False
+
+    with pytest.raises(RuntimeError):
+        Simulator(model)
```
tbekolay: I'm not sure if this makes a difference (since I believe the error happens in `__init__`) but in similar tests I've been creating the simulator as a context manager here with the thought that the `__exit__` might still fire and clean things up if anything was made.

nengo_loihi/loihi_cx.py
```diff
@@ -382,6 +382,11 @@ def get_loihi(self, seed=None):
     def get_simulator(self, seed=None):
         return CxSimulator(self, seed=seed)
 
+    def validate(self):
+        if len(self.cx_groups) == 0:
+            raise RuntimeError("No neurons marked for execution on-chip. "
```
tbekolay: I don't think we've been very judicious about this in Nengo Loihi yet, but we should be using subclasses of `NengoException` in order to make it easy for other tools to catch / deal with exceptions coming from Nengo separate from other exceptions. In this case, I think I would make this a `BuildError`.

nengo_loihi/tests/test_simulator.py
hunse: Sure, I'm fine with that.

nengo_loihi/loihi_cx.py
hunse: Sounds good!

hunse:
Ok, I've addressed your comments, @tbekolay. I also added a commit that replaces all the generic exceptions with nengo ones (except one! see the commit comment).

PR #39, opened 2018-08-31
keyword_spotting: proper amplitude now used

hunse:
This is the amplitude used during training. This increases the
accuracy of the network.

TODO: The amplitude is implemented manually (by multiplying with the transform). Using the `amplitude` parameter on the neuron_type does not work. It's not yet clear to me why. That could be fixed either here or in a future PR.


hunse:
Got amplitude working if pre is off-chip and post is on-chip. (This fixed the keyword spotting example.)

TODO: Need some tests for amplitude. The current test is not running since ReLUs aren't implemented. Will need further tests for the case that the pre-ensemble is off chip, and for the pre-ensemble on-chip and the post-ensemble off-chip.

tbekolay:
Are these TODOs things for this PR or for a future PR?

hunse:
I think they can wait for another PR, since my changes don't introduce any of these problems, just made me aware of them. I'll make an issue.

PR #41, opened 2018-09-04
WIP: Document and test hardware classes

hunse:
Document and unit-test the hardware-related classes. In particular, the allocator should be testable off-Loihi.

Based off #30.


tbekolay:
Making allocators into classes was done in another PR, so the only thing unique here are some docstrings. I rebased this to make that clear.

PR #42, opened 2018-09-04
Fix RectifiedLinear, and other fun fixes

hunse:
Got RectifiedLinear working by properly scaling the Vth.


nengo_loihi/builder.py
```diff
@@ -326,10 +326,11 @@ def build_lif(model, lif, neurons, group):
         dt=model.dt)
 
 
-# TODO: fix issues with ReLu before enabling it
-# @Builder.register(nengo.RectifiedLinear)
+@Builder.register(nengo.RectifiedLinear)
 def build_relu(model, relu, neurons, group):
-    group.configure_relu(dt=model.dt)
+    group.configure_relu(
+        vth=1./model.dt,  # so input == 1 -> neuron fires 1/dt steps -> 1 Hz
```
drasmuss: Are there any advantages/disadvantages to scaling the threshold versus scaling the input current?  I'm thinking of the scaling we do during discretization in particular, I know some of that is based on Vth.  Will changing Vth here affect the accuracy of the discretization or anything like that?

hunse: I think the discretization should end up the same either way. The reason I did it on the threshold is because that was easy to do in the neuron builder, whereas scaling the input would mean checking for RectifiedLinear in the ensemble builder and then scaling gain and bias (a bit messier).

tbekolay:
I'll test this out and merge if tests pass! I'll also add a changelog entry in the merge ... I think we don't need to be doing changelog entries for all (or even most) PRs, but for the ones that users might care about (like this one) I think we should do it as we go along so it doesn't have to be a big part of the release process.

tbekolay:
Ah yeah another thing about this PR... currently it uses `nengo.RectifiedLinear`, which also enables `nengo.SpikingRectifiedLinear`, so you can use either, but using `RectifiedLinear` will not match Nengo due to it being rate-based, while `SpikingRectifiedLinear` will match more closely. So should we only be allowing `SpikingRectifiedLinear` as it will more closely match Nengo?

hunse:
I would be okay with only allowing `SpikingRectifiedLinear`. That matches what we do with LIFs (i.e. we don't convert LIFRate to LIF).

PR #43, opened 2018-09-04
Do not overwrite filter

hunse:
With multiple connections into an ensemble, error if the the tau_s of one is different (and thus would overwrite) from that of the other.

Based on #42.

TODO: check that this doesn't break our existing recurrent networks (oscillator).

### DIFF
--- a/nengo_loihi/loihi_cx.py
+++ b/nengo_loihi/loihi_cx.py
@@ -83,6 +83,8 @@ class CxGroup(object):
         Whether these compartments are on a Loihi core
         or handled by the Loihi x86 processor (CPU).
     """
+    # threshold at which U/V scaling is allowed
+    DECAY_SCALE_TH = 0.5 / 2**12  # half of one decay scaling unit
 
     def __init__(self, n, label=None, location='core'):
         self.n = n
@@ -182,8 +184,12 @@ class CxGroup(object):
                 "Cannot change tau_s on already configured neurons")
 
         self.decayU[:] = decayU
-        self.scaleU = decayU > 1e-15
+        self.scaleU = decayU > self.DECAY_SCALE_TH
         self.decayU_set = not default
+        if not self.scaleU:
+            raise BuildError(
+                "Current (U) scaling is required. Perhaps a synapse time "
+                "constant is too large in your model.")
 
     def configure_lif(self, tau_rc=0.02, tau_ref=0.001, vth=1, dt=0.001):
         self.decayV[:] = -np.expm1(-dt/np.asarray(tau_rc))
@@ -191,7 +197,11 @@ class CxGroup(object):
         self.vth[:] = vth
         self.vmin = 0
         self.vmax = np.inf
-        self.scaleV = np.all(self.decayV > 1e-15)
+        self.scaleV = np.all(self.decayV > self.DECAY_SCALE_TH)
+        if not self.scaleV:
+            raise BuildError(
+                "Voltage (V) scaling is required with LIF neurons. Perhaps "
+                "the neuron tau_rc time constant is too large.")
 
     def configure_relu(self, tau_ref=0.0, vth=1, dt=0.001):
         self.decayV[:] = 0.
--- a/nengo_loihi/loihi_cx.py
+++ b/nengo_loihi/loihi_cx.py
@@ -27,14 +27,73 @@ class CxGroup(object):
 
     Typically an ensemble or node, can be a special decoding ensemble. Once
     implemented, SNIPS might use this as well.
+
+    Before ``discretize`` has been called, most parameters in this class are
+    floating-point values. Calling ``discretize`` converts them to integer
+    values inplace, for use on Loihi.
+
+    Attributes
+    ----------
+    n : int
+        The number of compartments in the group.
+    label : string
+        A label for the group (for debugging purposes).
+    decayU : (n,) ndarray
+        Input (synapse) decay constant for each compartment.
+    decayV : (n,) ndarray
+        Voltage decay constant for each compartment.
+    decayU_set : bool
+        Whether decayU has been explicitly set
+    scaleU : bool
+        Scale input (U) by decayU so that the integral of U is
+        the same before and after filtering.
+    scaleV : bool
+        Scale voltage (V) by decayV so that the integral of V is
+        the same before and after filtering.
+    refractDelay : (n,) ndarray
+        Compartment refractory delays, in time steps.
+    vth : (n,) ndarray
+        Compartment voltage thresholds.
+    bias : (n,) ndarray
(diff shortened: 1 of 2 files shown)


hunse:
Because ensemble-ensemble connections use interneurons, and the connection filtering is done by the interneuron population, this doesn't break the current oscillator examples.

This would stop any network that has e.g. multiple connections into a population's neurons (thus without interneurons) with different synapses. But of course, such examples are already broken, this is just catching that and stopping them from running.

drasmuss:
I was going to rebase this but just wanted to confirm; the "use mid_cx in connection builder" is supposed to be part of this PR right?

hunse:
I just went ahead and rebased it. I hope that's all right.

nengo_loihi/loihi_cx.py
```diff
@@ -101,17 +102,25 @@ def add_probe(self, probe):
         assert probe.target is self
         self.probes.append(probe)
 
-    def set_decay_U(self, tau_s, dt):
-        self.decayU[:] = 1 if tau_s == 0 else -np.expm1(-dt/np.asarray(tau_s))
+    def set_decay_U(self, tau_s, dt, default=False, overwrite=False):
```
tbekolay: Is `overwrite` something that we already use somewhere? Looking through the diff I don't see it being used, and because of that I would opt to not include this kwarg as I feel like it makes the function a bit harder to read. Also, it feels like if we end up needing to use it, we should instead refactor a bit so that we don't need to use.

tbekolay: The `default` logic is also a little confusing to me... it feels like if this was a normal default, I would initialize `decayU` as that default, and then use some sentinel value like `None` to say "hey I want to use default value here."

I can see that the default can't really be set because you use the first configured value. That's probably fine, but if there's a reasonable default value that we can use, we should use it.

Then, rather than having a separate `tau_s` and `default` kwargs, would it be possible to pass `tau_s=None` when you want to use the default? I feel like that would be less awkward... Alternatively, since using the default is a no-op if the group has already been configured, why don't we check for that outside of this method and not call this method if we don't need to configure?

nengo_loihi/loihi_cx.py
```diff
@@ -101,17 +102,25 @@ def add_probe(self, probe):
...
+        if default and self.decayU_set:
+            return
 
-    def configure_filter(self, tau_s, dt=0.001):
-        """Synaptic input filter for Cx."""
+        decayU = 1 if tau_s == 0 else -np.expm1(-dt/np.asarray(tau_s))
+        if not overwrite and self.decayU_set and not np.allclose(
+                decayU, self.decayU):
+            raise ValueError(
```
tbekolay: I would make this a `BuildError` or `ValidationError` (not sure when this error might be raised, but I suspect it's during the build).

nengo_loihi/loihi_cx.py
```diff
@@ -101,17 +102,25 @@ def add_probe(self, probe):
...
-        self.set_decay_U(tau_s, dt)
+    def configure_filter(self, tau_s, dt=0.001, **kwargs):
+        """Synaptic input filter for Cx."""
+        self.set_decay_U(tau_s, dt, **kwargs)
 
-    def configure_lif(
-            self, tau_s=0.005, tau_rc=0.02, tau_ref=0.001, vth=1, dt=0.001):
-        self.set_decay_U(tau_s, dt)
+    def configure_lif(self, tau_rc=0.02, tau_ref=0.001, vth=1, dt=0.001):
+        self.set_decay_U(0.005, dt, default=True)
```
tbekolay: What does `tau_s` represent? I don't think it's documented anywhere but documenting it would help in determining whether it makes sense to hard code it for LIF and ReLu.

nengo_loihi/loihi_cx.py
hunse: Those are fair points. I had included it just because I generally don't like methods that don't let me do what I want even if I'm saying "this is what I want". I sometimes find that causes me to hack things by just looking at the internals to the method and calling them manually. But I do agree that we should hopefully be able to not overwrite, and if we do run into a sticky situation we can always figure it out later. I'll take it out.

hunse: > since using the default is a no-op if the group has already been configured, why don't we check for that outside of this method and not call this method if we don't need to configure?

I think that's an option, and if you think that's cleaner I'm fine to change. I felt that doing it the way I did was cleaner, because it kept the `decay_U_set` parameter internal (so all interaction can happen through the `set_decay_U` function). But I'm not strongly opinionated on that.

The reason we can't just set the default up top when we create `decayU` is that I feel like the default should be 5 ms, but to figure out the corresponding decay value, you need to know `dt`. Also, it seems like that default is more associated with the builder (where we're setting up the interneurons, since I essentially use it as a default for interneuron connections).

You can't pass `tau_s=None` to use the default, because the point of `default=True` is that you're setting the default (so you need to pass `tau_s` as well).

nengo_loihi/loihi_cx.py
hunse: Will do

nengo_loihi/loihi_cx.py
hunse: It's the synapse time constant on the neuron input. It's not hard-coded here (`default=True`), but I agree, it should be documented. The default actually probably doesn't have to be set here at all, and in fact maybe it shouldn't.

nengo_loihi/loihi_cx.py
tbekolay: I agree in principle that you don't want to look at places other than the method to do what you want, but in this case it's only being used internally anyhow so we can expect some familiarity with the class. In any case, we can revisit later if we run into situation where we really want to overwrite.

hunse:
Ok, I think I've addressed all comments, except the one about getting rid of the `default` parameter (I have documented it instead).

I also added a commit that makes sure `decayU` is not zero once we discretize. This could happen when using very long synapse time constants. In such a case, there is no good value for `scaleU` since either it is True, in which case we are scaling voltage thresholds etc. up by a huge factor, or False, in which case there is no scaling and the filtered value will be *much* larger than expected.

hunse:
Yeah, I was just trying to save space :) Especially since there are so many attributes and I think people will mostly be reading that documentation in code (i.e. it's not for end users). But I guess we should do it properly. And once we do the refactoring, then each file will have less in it, so I don't mind longer documentation as much.

tbekolay:
> Yeah, I was just trying to save space

Makes sense, the numpydoc parsing depends on the newline unfortunately. Want me to fix it in the merge?

hunse:
I've fixed it. So if it looks good, go ahead and merge.

Issue #8, opened 2018-08-20
Only add encoders if they're used


tbekolay:
As suggested by @hunse: in 8a1420ff9769eb98d3a we can now have ensembles with only connections into their neurons. This means that the encoders are not used. However, the unused encoders will still affect how the weight and vth scaling is chosen in `discretize`. It is probably best to remove them if they're not used (or rather don't add them until a connection in the builder specifically asks for them), both for better weight scaling, and so they don't take up space on the chip.

hunse:
Fixed in #44.

PR #44, opened 2018-09-04
Encoders on demand

hunse:
Only build the encoders if they're needed. Addresses #8.

Based on #43.


nengo_loihi/loihi_cx.py
```diff
@@ -252,8 +252,9 @@ def discretize(target, value):
 
 
 class CxSynapses(object):
-    def __init__(self, n_axons):
+    def __init__(self, n_axons, label=None):
```
tbekolay: I like adding the labels a lot, makes the code that creates `CxSynapses` more clear :+1:

nengo_loihi/builder.py
```diff
@@ -289,21 +289,9 @@ def build_ensemble(model, ens):
...
-    synapses2 = CxSynapses(2*scaled_encoders.shape[1], label="inter_encoders")
-    inter_scale = 1. / (model.dt * INTER_RATE * INTER_N)
-    interscaled_encoders = scaled_encoders * inter_scale
-    synapses2.set_full_weights(
-        np.vstack([interscaled_encoders.T, -interscaled_encoders.T]))
-    group.add_synapses(synapses2, name='encoders2')
-
     model.add_group(group)
 
+    model.objs[ens]['group'] = group
```
tbekolay: It feels a little weird that we're storing the group 4 separate times ... is there a possibility that the `'in'` and `'out'` would be something else?

nengo_loihi/builder.py
```diff
@@ -651,6 +652,9 @@ def build_connection(model, conn):
         if conn.learning_rule_type is not None:
             raise NotImplementedError()
     elif isinstance(conn.post_obj, Ensemble):
+        if 'encoders2' not in post_cx.named_synapses:
```
tbekolay: Could we use a more descriptive name than `'encoders2'`?

nengo_loihi/builder.py
hunse: I didn't want to assume, but you're probably right. I'll change it.

PR #45, opened 2018-09-04
Always set filter on post ensemble in connection

hunse:
This would fix (for example) neuron->ensemble connections
(of which we currently have no test cases).

Unfortunately, this changes the behaviour of node->ensemble connections,
causing node_ens_ens to fail. The synapse set on the connection (None)
overrides the default INTER_TAU synapse, removing the filtering on
the ReLU neurons used to turn the node values into spikes.

Based on #43.

### DIFF
--- a/docs/examples/node_ens_ens.py
+++ b/docs/examples/node_ens_ens.py
@@ -23,7 +23,7 @@ with nengo.Network(seed=seed) as model:
     up = nengo.Probe(u, synapse=None)
 
     a = nengo.Ensemble(100, d, label='a')
-    nengo.Connection(u, a, synapse=None)
+    nengo.Connection(u, a)
     ap = nengo.Probe(a)
     anp = nengo.Probe(a.neurons)
     avp = nengo.Probe(a.neurons[:5], 'voltage')
--- a/docs/examples/node_ens_ens.py
+++ b/docs/examples/node_ens_ens.py
@@ -18,7 +18,7 @@ bnp = None
 nengo_loihi.set_defaults()
 with nengo.Network(seed=seed) as model:
     u = nengo.Node(
-        output=nengo.processes.WhiteSignal(tend, high=5, seed=seed + 1),
+        output=nengo.processes.WhiteSignal(tend, high=2, seed=seed + 1),
         size_out=d)
     up = nengo.Probe(u, synapse=None)
 
--- a/nengo_loihi/builder.py
+++ b/nengo_loihi/builder.py
@@ -305,6 +305,7 @@ def build_ensemble(model, ens):
         group.bias[:] = bias
         model.build(ens.neuron_type, ens.neurons, group)
 
+    # set default filter just in case no other filter gets set
     group.configure_filter(INTER_TAU, dt=model.dt, default=True)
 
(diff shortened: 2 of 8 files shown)


hunse:
Based on the discussion in #63, I think this is ready for review.

hunse:
@tcstewar: I've added some comments. Let me know what you think.

One situation that we're going to have trouble with in general is that each group of neurons can only have one input synapse. So right now, in the case where we have a recurrent connection (which would have a INTER_TAU filter going back in to the neurons) and an input node (with this change, would now apply its own filter), things would break.

One solution, especially since it seems we don't *need* filtering on the interneurons (at least on node interneurons, I haven't tested on ensemble ones where we use 20 neurons at 100 Hz instead of 2 at 1000 Hz), would be to only use INTER_TAU as a default for ensembles, and if anything else overrides that, then that would be the new input filter.

hunse:
So I think I'd added the comments, but not pushed them. Sorry @tcstewar! If you want to take a look now, go ahead.

I think @tbekolay is looking to merge this soon. I was trying to figure out what I was talking about in that last comment about only making INTER_TAU the default for ensembles; it seems like that was done in cb6ba75. As far as I can tell, this is good to go.

tcstewar:
> So I think I'd added the comments, but not pushed them. Sorry @tcstewar! If you want to take a look now, go ahead.

Those comments look good to me!  Thank you for clarifying!  :)

tbekolay:
After rebasing, this fails `test_examples::test_node_ens_ens` and `test_ensemble::test_amplitude`. Are these failures a result of the rebase or were they failing before? How to fix?

hunse:
Fails in emulator or on Loihi?

hunse:
I just rebased myself and found this failure in the emulator for test_node_ens_ens. It did not fail before the rebase (in the emulator). I'll look into these failures.

tbekolay:
Emulator, therefore probably both. You can see the failures in [travis CI](https://travis-ci.com/nengo/nengo-loihi/jobs/147577657).

drasmuss:
Could we add a warning for node->* connections with `synapse=None`?  It'll definitely surprise people that synaptic filters are necessary on Nodes.

tbekolay:
@hunse Could you explain a bit what the issue is with having no filter on nodes? Is it because node inputs are translated to spikes later on? Is there a mechanism for injecting current? Obviously not things to look into in this PR, but a bit of this explanation would be good to have on the warning raised for node->* connections.

hunse:
So yeah, all node output gets turned into spikes by this ensemble that we introduce in the splitter: https://github.com/nengo/nengo-loihi/blob/619e45bf00305b758e030cd4ddb40829a09b382d/nengo_loihi/splitter.py#L187

The filter on the node connection is applied to the output spikes of that ensemble. Things work without a filter, they're just a little noisier.

Another option for down the road would be to have the standard `INTER_TAU` filter on the output of that ensemble (on the chip), and apply our connection filter on the connection going into that ensemble.

hunse:
We're not good about accounting for synaptic delays in our tests, so I'm
guessing that's the cause of the RMSE change.

PR #46, opened 2018-09-05
NotImplementedError for ensemble noise

hunse:
Before, this would just silently fail.


hunse:
Yeah, that is the state of things. The thing is that it's really difficult to actually map our noise to their noise. On their noise, we can only set the exponent. Thus the magnitude has to be in powers of two. But you're right, we do have the lower-level logic set up to turn noise on on the board.

PR #48, opened 2018-09-05
Enable running Nengo tests

tbekolay:
This adds some additional options to `pytest.ini` which makes it possible to run the Nengo tests with

```bash
pytest --pyargs nengo
```

Most of the test failures I'm seeing right now (just running on the emulator) are due to issues already fixed in existing pull requests, so the number of passing tests isn't very meaningful.

Still todo:

- [x] Figure out how to deal with tests with no ensembles
- [x] Expand `allclose` fixture to deal with different tolerances on different backends (@drasmuss)
- [x] Reenable SpikingReLu tests
- [x] Make sure all sims are properly closed (in tests, and in `nengo_loihi/simulator.py`)

[Additional discussion](https://gitter.im/nengo/loihi-48)

### DIFF
--- a/nengo_loihi/builder.py
+++ b/nengo_loihi/builder.py
@@ -6,6 +6,7 @@ import numpy as np
 
 import nengo
 from nengo import Network, Ensemble, Connection, Node, Probe
+from nengo.builder.connection import BuiltConnection
 from nengo.dists import Distribution, get_samples
 from nengo.connection import LearningRule
 from nengo.ensemble import Neurons
@@ -134,6 +135,12 @@ class Model(CxModel):
         averaged over time will be 1.
         """
         return 1. / (self.dt * self.inter_rate * self.inter_n)
+
+    def __getstate__(self):
+        raise NotImplementedError("Can't pickle nengo_loihi.builder.Model")
+
+    def __setstate__(self, state):
+        raise NotImplementedError("Can't pickle nengo_loihi.builder.Model")
 
     def __str__(self):
         return "Model: %s" % self.label
@@ -326,12 +333,9 @@ def build_ensemble(model, ens):
     gain, bias, max_rates, intercepts = get_gain_bias(
         ens, rng, model.intercept_limit)
 
-    if isinstance(ens.neuron_type, nengo.Direct):
-        raise NotImplementedError()
-    else:
-        group = CxGroup(ens.n_neurons, label='%s' % ens)
-        group.bias[:] = bias
-        model.build(ens.neuron_type, ens.neurons, group)
+    group = CxGroup(ens.n_neurons, label='%s' % ens)
+    group.bias[:] = bias
+    model.build(ens.neuron_type, ens.neurons, group)
 
     # set default filter just in case no other filter gets set
     group.configure_default_filter(model.inter_tau, dt=model.dt)
@@ -376,6 +380,18 @@ def build_interencoders(model, ens):
     group.add_synapses(synapses, name='inter_encoders')
 
 
+@Builder.register(nengo.neurons.NeuronType)
+def build_neurons(model, neurontype, neurons, group):
+    # If we haven't registered a builder for a specific type, then it cannot
+    # be simulated on Loihi.
+    raise BuildError(
+        "The neuron type %r cannot be simulated on Loihi. Please either "
+        "switch to a supported neuron type like LIF or "
+        "SpikingRectifiedLinear, or explicitly mark ensembles using this "
+        "neuron type as off-chip with\n"
+        "  net.config[ensembles].on_chip = False")
+
+
 @Builder.register(nengo.LIF)
 def build_lif(model, lif, neurons, group):
     group.configure_lif(
@@ -400,11 +416,6 @@ def build_node(model, node):
         return
     else:
         raise NotImplementedError()
-
-
-BuiltConnection = collections.namedtuple(
-    'BuiltConnection',
-    ('eval_points', 'solver_info', 'weights', 'transform'))
 
 
 def get_eval_points(model, conn, rng):
@@ -534,7 +545,7 @@ def build_connection(model, conn):
             # TODO: this identity transform may be avoidable
             transform = np.eye(conn.pre.size_out)
         else:
-            assert transform.ndim == 2
+            assert transform.ndim == 2, "transform shape not handled yet"
             assert transform.shape[1] == conn.pre.size_out
 
         assert transform.shape[1] == conn.pre.size_out
@@ -565,7 +576,7 @@ def build_connection(model, conn):
             needs_interneurons = True
     elif isinstance(conn.pre_obj, Neurons):
         assert conn.pre_slice == slice(None)
-        assert transform.ndim == 2
+        assert transform.ndim == 2, "transform shape not handled yet"
         weights = transform / model.dt
         neuron_type = conn.pre_obj.ensemble.neuron_type
     else:
--- a/nengo_loihi/simulator.py
+++ b/nengo_loihi/simulator.py
@@ -44,7 +44,7 @@ class ProbeDict(NengoProbeDict):
                 if key in fallback:
                     target = fallback.raw
                     break
-        assert key in target
+        assert key in target, "probed object not found"
 
         if (key not in self._cache
                 or len(self._cache[key]) != len(target[key])):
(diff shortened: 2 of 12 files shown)


drasmuss:
I was thinking I could start working on that flexible `allclose` fixture we were talking about, if you haven't started on that already?

tbekolay:
I haven't started on it, so go ahead! I'll consider this ready for collaboration then, and start a todo list on the main post.

drasmuss:
`allclose` fixture is implemented (https://github.com/nengo/nengo/commits/allclose), so next step I'll work on is to go through the tests and see which need their tolerances adjusted.

drasmuss:
Note: this depends on https://github.com/nengo/nengo/pull/1477

hunse:
tests failing due to the "300" bug now work with the newest NxSDK, since they fixed that bug. They should be re-enabled here.

tbekolay:
This PR is now ready for review. This is now set up so that we run the Nengo tests during CI, which means we can track when Nengo tests start failing or unexpectedly passing. I'm not sure if we can do this automatically, but we should at least manually check in PRs if any Nengo tests result in an `xpass` because it would be good to track which commits result in passing more Nengo tests. The notebook that was in here previously is now in the sandbox.

Performance, in terms of the the number of tests passing, is slightly diminished from the original PR because this PR does not rely on the `allclose` branch of Nengo anymore. I have separated the changes that depend on that branch so that we can incorporate them later once we've done a release of `pytest-allclose`. Until then, though, the important thing is that we're regularly running the Nengo tests and tracking when we start passing more / less. Our percentage passing will go up quickly now that we're tracking it, I suspect.

@hunse could you review this? Also, I forgot to pull this branch before making changes, so I lost your commits modifying the list of hanging tests on hardware (sorry!!) I suspect you have those commits locally, so please add them to this branch as part of your review.

nengo_loihi/builder.py
```diff
@@ -326,12 +333,12 @@ def build_ensemble(model, ens):
     gain, bias, max_rates, intercepts = get_gain_bias(
         ens, rng, model.intercept_limit)
 
-    if isinstance(ens.neuron_type, nengo.Direct):
-        raise NotImplementedError()
-    else:
+    if isinstance(ens.neuron_type, (nengo.LIF, nengo.SpikingRectifiedLinear)):
```
hunse: This prohibits someone from adding their own build hook for a neuron type. Can we leave this as before? The builder should raise an error if it doesn't find a build hook for the neuron type in question.

drasmuss:
> I'm not sure if we can do this automatically, but we should at least manually check in PRs if any Nengo tests result in an xpass because it would be good to track which commits result in passing more Nengo tests. 

We can set the xfail_strict=True option, which treats xpass as a failure.

nengo_loihi/simulator.py
```diff
@@ -141,16 +141,173 @@ class Simulator(object):
...
+        # utils.connection.target_function (deprecated)
+        ('utils/tests/test_connection.py*',
+         "target_function (deprecated) not working"),
+    ]
 
     def __init__(self, network, dt=0.001, seed=None, model=None,  # noqa: C901
-                 precompute=False, target=None):
+                 precompute=False, target=None, progress_bar=None):
         self.closed = True  # Start closed in case constructor raises exception
+        self._time = 0
```
hunse: Where is this used?

nengo_loihi/splitter.py
```diff
@@ -154,8 +154,9 @@ def place_ensembles(networks):
         # User-specified config takes precedence
         if config[ens].on_chip is not None:
             networks.move(ens, "chip" if config[ens].on_chip else "host")
-        # Direct mode ensembles must be off chip
-        elif isinstance(ens.neuron_type, nengo.Direct):
+        # Unsupported neuron types must be off-chip
+        elif not isinstance(
+                ens.neuron_type, (nengo.LIF, nengo.SpikingRectifiedLinear)):
```
hunse: This is an interesting problem. On the one hand, I like that someone's model will just work. On the other hand, I don't like that an ensemble with a different neuron type will be silently moved off-chip. I feel like this quiet splitting is what's led to problems in the past (with us thinking more of the keyword spotter was on chip than actually was), and will likely confuse users as well (if I pass my model to `nengo_loihi`, I kind of assume it's going to run it almost exclusively on Loihi).

Also, it runs into the same problem as I pointed out above, where people can't add hooks for other neuron types. The way things are currently set up, this actually might not be as big of a problem as I was thinking because a) people can subclass off `LIF` or `SpikingRectifiedLinear` and b) there aren't really any single-compartment models that Loihi supports other than these two. However, as we open the door to multi-compartment, there may be either multi-compartment models that someone may want to add, or a way of doing a more complicated single-compartment model using multiple Loihi compartments.

drasmuss: I agree that silently simulating these neurons off-chip will probably lead to confusion.  But it would also be nice to support, so that if someone really does intend to build a mixed model they can.  Maybe we could make this an explicit opt-in somehow? Something like
```
net.config[nengo.Izhikevich].on_chip = False
```
(building on our existing `on_chip` logic for ensembles).

tbekolay:
> We can set the xfail_strict=True option, which treats xpass as a failure.

Oh nice, I'll do that :+1:

nengo_loihi/builder.py
tbekolay: Yeah I think we can leave this as before as the issue was more with the splitter allowing you to get to this point with neuron types we didn't expect. By the same token, since we don't implement a builder to `nengo.Direct` do we even need this block?

nengo_loihi/simulator.py
tbekolay: In the `time` property. If you create a `Simulator` instance but don't call `run`, doing `sim.time` gives an `AttributeError`.

nengo_loihi/splitter.py
tbekolay: I guess we could have a registry of known neuron types that is based on the builder? That is I think quick enough that we should do it in this PR ... the problem of things not happening on chip that you think are happening on chip is a tough one, like should we raise a warning for every ensemble that won't be put on chip? Right now there's an exception if all ensembles are off-chip, which will hopefully be sufficient for the majority of cases where people think Loihi does something that it doesn't. But I think that particular issue should be deferred to another PR so as to not block the release, I'll implement the neuron type registry though so people can add other builders.

hunse: Do we need this change for the nengo tests to work? Or is it an option to keep going as we were, with Ensembles that have a neuron type that is not supported on Loihi just raising an error? The option to put individual ensembles off-chip is of course there, so if someone has an ensemble with a different neuron type they can manually specify `on_chip = False`. It also wouldn't be hard for a user to loop over all ensembles in a model and set `on_chip = False` to any that have a specific neuron type, or don't have `LIF` or `SpikingRectifiedLinear`.

nengo_loihi/simulator.py
hunse: Ahh, I see, we had this attribute before, it just was not initialized. I think it should be initialized in `reset` where `_n_steps` is set to zero. Does that make sense?

nengo_loihi/builder.py
hunse: Yeah, I was wondering about that last part too. Probably not. I just copied the builder from Nengo and modified it, so this is just a remnant of that.

hunse: Added a fixup for this.

nengo_loihi/simulator.py
hunse: Added a fixup for this.

nengo_loihi/splitter.py
tbekolay: It's not strictly required, the question is how you interpret a particular test failure. With this change, these failing tests fall in the "there are no ensembles being simulated on-chip" category, which is pretty clearly something that we are not planning to fix in Nengo Loihi. Before, the test failed with "hey there's no builder for this neuron type," which looks like something that we just haven't implemented yet.

One alternative would be to add builders for these neuron types that explicitly say "this neuron type cannot be simulated on Loihi. Change the neuron type, or mark this ensemble as off chip with `net.config[ens].on_chip = False`. From a user perspective, though, I would think "why not just detect that this can't run on Loihi and do it automatically?" So it's a balance between convenience and explicitness.

We talked in the past about having some easy way to print out a summary of what's on the chip and how it's organized. I feel like when we have that, we'll want this behavior, but perhaps until then we want to go the explicit route. But I think the explicit route includes the explicit error I mention in the previous paragraph.

drasmuss: :+1: for explicit error now, and then handling it automatically once we have a better way to convey information about what is running on/off chip.

tbekolay:
> > We can set the xfail_strict=True option, which treats xpass as a failure.
>
> Oh nice, I'll do that ￼:+1:

I did that, but we have several tests in the main Nengo Loihi test suite that xpass on the emulator but not hardware and that kind of thing, so I left this as a separate issue, #144.

nengo_loihi/splitter.py
tbekolay: Okay, I reverted this change and added the build error in 8f202dd64fd4b89a3ee400dc27ec8f492e3a67e7.

hunse: I like that fix. For me, I guess I don't expect `nengo_loihi` to do a lot of moving stuff around for me to get my model to work. If I have a neuron model that doesn't work on Loihi, I'm more apt to change that neuron model than to move the ensemble off-chip. I like that this change makes the decision clear to the user, and makes them be explicit about whatever decision they make.

hunse:
Pushed a fixup that should fix the failing `test_sigmoid_invalid`.

PR #49, opened 2018-09-05
Add more accurate LIF rate function for decoders

hunse:
This rate function accounts for Loihi discretization.

TODO: I'd like to look qualitatively/quantitatively at how much effect this actually has on some more integrated test cases.

Just looking at `test_connection.py`, I found that `mean rmse: 0.99851 +/- 1.8416` before the change and `mean rmse: 0.90317 +/- 1.7610` after. So there's some improvement.

TODO: Once the ReLU fix is merged, I'll rebase and do the same for that.

### DIFF
--- a/nengo_loihi/neurons.py
+++ b/nengo_loihi/neurons.py
@@ -8,6 +8,39 @@ from nengo.neurons import NeuronType
 from nengo.params import NumberParam
 
 
+def loihi_lif_rates(neuron_type, x, gain, bias):
+    dt = 0.001
+    j = neuron_type.current(x, gain, bias) - 1
+
+    out = np.zeros_like(j)
+    period = neuron_type.tau_ref + neuron_type.tau_rc * np.log1p(1. / j[j > 0])
+    out[j > 0] = (neuron_type.amplitude / dt) / np.ceil(period / dt)
+    return out
+
+
+def loihi_spikingrectifiedlinear_rates(neuron_type, x, gain, bias):
+    dt = 0.001
+    j = neuron_type.current(x, gain, bias)
+
+    out = np.zeros_like(j)
+    period = 1. / j[j > 0]
+    out[j > 0] = (neuron_type.amplitude / dt) / np.ceil(period / dt)
+    return out
+
+
+def loihi_rates(neuron_type, x, gain, bias):
+    for cls in type(neuron_type).__mro__:
+        if cls in loihi_rate_functions:
+            return loihi_rate_functions[cls](neuron_type, x, gain, bias)
+    return neuron_type.rates(x, gain, bias)
+
+
+loihi_rate_functions = {
+    nengo.LIF: loihi_lif_rates,
+    nengo.SpikingRectifiedLinear: loihi_spikingrectifiedlinear_rates,
+}
+
+
 class NIFRate(NeuronType):
     """Non-spiking version of the non-leaky integrate-and-fire (NIF) model.
 
--- a/nengo_loihi/neurons.py
+++ b/nengo_loihi/neurons.py
@@ -8,18 +8,18 @@ from nengo.neurons import NeuronType
 from nengo.params import NumberParam
 
 
-def loihi_lif_rates(neuron_type, x, gain, bias):
-    dt = 0.001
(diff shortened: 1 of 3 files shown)


hunse:
Ok, done for `SpikingRectifiedLinear` now too.

hunse:
Looks good! I've added the [`neuron-dt` branch](https://github.com/nengo/nengo-loihi/commits/neuron-dt) to start looking at changing the `dt` (these curves do not seem to work for other `dt`).

The question is, whether we want to merge this as-is and make that an issue, or try to track that down now. @tcstewar @tbekolay opinions?

tbekolay:
I feel like we haven't tested with other dts at all, I would raise an error if `dt != 0.001` for now and do a thorough check of dt issues in another PR.

tbekolay:
Is there anything else you wanted to do on this @tcstewar? If not, please assign back to @hunse.

hunse:
Ok. I did just get this working for other dts, so maybe I'll put that code here. But I agree, it would be good to error (or at least warn) right now for other dts, since it hasn't been well tested.

hunse:
I added a commit outlawing `dt != 0.001`. If one of you can sign off on that, I think this is good to go.

I'll make an issue related to other dt.

tcstewar:
> I added a commit outlawing dt != 0.001. If one of you can sign off on that, I think this is good to go.

Yup, that looks good to me!  I do think that other issue (about getting dt!=0.001 working) would be great, but it makes sens to error out if we haven't confirmed it....

hunse:
See issue #54 and PR #55 for the continuation of the `dt` stuff.

nengo_loihi/neurons.py
```diff
@@ -8,6 +8,41 @@
...
 
 
+def loihi_lif_rates(neuron_type, x, gain, bias, dt):
+    # discretize tau_ref as per CxGroup.configure_lif
+    tau_ref = dt * np.round(neuron_type.tau_ref / dt)
+
+    j = neuron_type.current(x, gain, bias) - 1
+
+    out = np.zeros_like(j)
+    p = tau_ref + neuron_type.tau_rc * np.log1p(1. / j[j > 0])
```
tbekolay: What is `p`? In generally I'm really not a big fan of one-letter variable names as they're hard to grep for, though I am ok for well known names like `j` for current.

nengo_loihi/neurons.py
```diff
@@ -8,6 +8,41 @@
...
+def loihi_spikingrectifiedlinear_rates(neuron_type, x, gain, bias, dt):
+    j = neuron_type.current(x, gain, bias)
+
+    out = np.zeros_like(j)
+    p = 1. / j[j > 0]
+    out[j > 0] = (neuron_type.amplitude / dt) / np.ceil(p / dt)
+    return out
+
+
+def get_loihi_rates(neuron_type, x, gain, bias, dt):
```
tbekolay: Could this be `loihi_rates`? I don't think the `get` adds to understanding what the function does.

nengo_loihi/tests/test_neurons.py
```diff
@@ -0,0 +1,45 @@
...
+        ap = nengo.Probe(a.neurons)
+
+    dt = 0.001
+    with Simulator(model, dt=dt) as sim:
+        sim.run(1.0)
+
+    est_rates = sim.data[ap].mean(axis=0)
+    ref_rates = get_loihi_rates(neuron_type, x, gain, bias, dt=dt)
+
+    # plt.plot(x, ref_rates, "k", label="predicted")
```
tbekolay: OK to remove these?

nengo_loihi/neurons.py
hunse: It's the period.

nengo_loihi/neurons.py
hunse: Sure

nengo_loihi/tests/test_neurons.py
hunse: I actually bring them back in #55. They're the better way to do it, because they work even if part of `x` is negative. The `semilog` just helps because so much of the interesting LIF stuff is crammed into the lower part of the curve. But really I think we should go back to just using `plot`.

Issue #54, opened 2018-09-10
Allow other values of `dt`


hunse:
We currently only allow `dt == 0.001`, since this is the Nengo default.

Most (all?) of the code has been written with the intention of supporting other values of `dt`, but this functionality has not yet been tested. We need to add some tests to be confident that the simulator works with other values of `dt`, then we can allow it.

hunse:
Done in 4b59936

PR #55, opened 2018-09-10
Allow `dt` other than 0.001

hunse:
Currently, this PR begins the process of testing other values of `dt`, so that we can enable this functionality.

We should add other `dt` tests here to get this fully tested.

Addresses #54.

### DIFF
--- a/nengo_loihi/splitter.py
+++ b/nengo_loihi/splitter.py
@@ -1,6 +1,7 @@
 import logging
 
 import nengo
+from nengo.exceptions import BuildError
 import numpy as np
 
 from nengo_loihi import loihi_cx
@@ -92,7 +93,7 @@ class ChipReceiveNeurons(ChipReceiveNode):
         super(ChipReceiveNeurons, self).__init__(dimensions, dimensions)
 
 
-def split(model, inter_rate, inter_n):  # noqa: C901
+def split(model, inter_rate, inter_n, dt):  # noqa: C901
     """Split a model into code running on the host and on-chip"""
 
     logger.info("Splitting model into host and chip parts")
@@ -181,7 +182,10 @@ def split(model, inter_rate, inter_n):  # noqa: C901
                     nengo.Connection(receive, c.post, synapse=c.synapse)
                 with host:
                     max_rate = inter_rate * inter_n
-                    assert max_rate <= 1000
+                    if max_rate > (1. / dt):
+                        raise BuildError(
+                            "Simulator `dt` must be <= %s (got %s)" % (
+                                1. / max_rate, dt))
 
                     logger.debug("Creating NIF ensemble for %s", c)
                     ens = nengo.Ensemble(
--- a/nengo_loihi/splitter.py
+++ b/nengo_loihi/splitter.py
@@ -182,7 +182,8 @@ def split(model, inter_rate, inter_n, dt):  # noqa: C901
                     nengo.Connection(receive, c.post, synapse=c.synapse)
                 with host:
                     max_rate = inter_rate * inter_n
-                    if max_rate > (1. / dt):
+                    rtol = 1e-8  # allow for floating point inaccuracies
+                    if max_rate > (1. / dt) * (1 + rtol):
                         raise BuildError(
                             "Simulator `dt` must be <= %s (got %s)" % (
                                 1. / max_rate, dt))
--- a/nengo_loihi/tests/test_simulator.py
+++ b/nengo_loihi/tests/test_simulator.py
@@ -1,4 +1,5 @@
 import nengo
+import numpy as np
 import pytest
 
 import nengo_loihi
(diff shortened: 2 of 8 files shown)


tcstewar:
This would be awesome!  :)

As I mentioned in a different PR, here's a random bit of the code that might be assuming dt=0.001, but I haven't had a chance to look into it in detail, so I thought I'd just note it here:

https://github.com/nengo/nengo-loihi/blob/master/nengo_loihi/splitter.py#L184

```python
                    max_rate = inter_rate * inter_n
                    assert max_rate <= 1000
```

drasmuss:
Generally speaking we want to rebase the branch onto master and run the unit tests before approving a PR (so that we're sure we're not breaking master, rather than maybe breaking master and then fixing that in future PRs).  I can do that though if you like.

drasmuss:
I went ahead and rebased/tested it since I was here anyway, but I'm getting a test failure in `test_neurons.py:test_loihi_rates` (on emulator, haven't tried hardware yet).  There was a merge conflict I had to resolve so I might have done that incorrectly, or there might be some new change in master that this doesn't work with.  Going to assign it back to @hunse.

tcstewar:
> Generally speaking we want to rebase the branch onto master and run the unit tests before approving a PR (so that we're sure we're not breaking master, rather than maybe breaking master and then fixing that in future PRs).

Ah, good point.  I will try to be better about that.  Thank you for going ahead and taking a stab at it!

nengo_loihi/tests/test_simulator.py
```diff
@@ -46,7 +47,11 @@ def test_dt(dt, a_on_chip, Simulator, seed, plt, allclose):
         nengo.Connection(a, b, function=function,
                          solver=nengo.solvers.LstsqL2(weights=True))
 
-    with Simulator(model, dt=dt, precompute=False) as sim:
+    loihi_model = nengo_loihi.builder.Model(dt=dt)
+    if dt > 1. / (loihi_model.inter_rate * loihi_model.inter_n):
+        loihi_model.inter_rate = 1. / (dt * loihi_model.inter_n)
```
drasmuss: Should this be something we're doing in all models (adjusting inter_rate for larger dts)?  That's not something that users are likely to know how to do, so if it's required to use dt>0.001 we should do it automatically, or if that isn't possible then raise a warning telling users how to adjust inter_rate.

nengo_loihi/conftest.py
```diff
@@ -111,11 +111,18 @@ def _allclose(a, b, rtol=1e-5, atol=1e-8, equal_nan=False):
 
 @pytest.fixture
 def assert_allclose(request):
-    def _assert_allclose(a, b, rtol=1e-5, atol=1e-8, equal_nan=False):
+    def _assert_allclose(a, b, rtol=1e-5, atol=1e-8, xtol=0, equal_nan=False):
```
drasmuss: Once this fixture is using `allclose`, it'd be good to have the `xtol` implementation in `allclose` instead (since that seems like something that would be useful in lots of places).

nengo_loihi/conftest.py
```diff
@@ -107,3 +107,20 @@ def _allclose(a, b, rtol=1e-5, atol=1e-8, equal_nan=False):
             request.node.user_properties.append(("rmse", rmse))
         return np.allclose(a, b, rtol=rtol, atol=atol, equal_nan=equal_nan)
     return _allclose
+
+
+@pytest.fixture
+def assert_allclose(request):
```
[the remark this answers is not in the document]
hunse: Yeah, I had made a note to do that in #67, but we can do it here, too.

nengo_loihi/tests/test_simulator.py
hunse: In that case, I would just set `inter_rate` to be `1. / (dt * loihi_model.inter_n)` by default. Maybe a sensible thing to do.

nengo_loihi/conftest.py
drasmuss: Also a test/documentation for the `xtol` behaviour would be great, since it's not obvious and the implementation is a bit complicated.

hunse:
Ok, I think I've addressed all your comments @drasmuss. I'm not quite sure how to add documentation to fixtures, so I didn't take the time right now to figure that out for `xtol`, but I did add an inline comment that hopefully makes it slightly clearer.

hunse:
Ok, those tests should be fixed. It was just a floating point rounding error that made the actual max_rate in splitter.py slightly too large.

drasmuss:
I tried to rebase this to `master`, but when I do that I get an assertion failure in `test_examples:test_node_ens_ens`.  There were a lot of conflicts so it's possible that something got messed up there (although nothing obvious I can see).  So I think it's some more subtle interaction between the `dt` changes and something else that has been merged in (or perhaps the upgrade to 0.7).

Pushed the rebased version to `neuron-dt-rebase` if you want to take a look.

hunse:
Ok, got it rebased.

drasmuss:
I'm still seeing the same error (this is on Loihi).

tbekolay:
@hunse Any progress on that error? We were hoping to have this in the release today, if I recall correctly.

hunse:
So I can't get the test (`test_examples:test_node_ens_ens`) to fail on Loihi, either with this branch or the one that @drasmuss rebased (`neuron-dt-rebase`). I ran it a number of times on each branch. @tbekolay, do you want to try to get it to fail, slash run all the tests?

nengo_loihi/splitter.py
```diff
@@ -92,7 +94,7 @@ def __init__(self, dimensions, neuron_type=None):
         super(ChipReceiveNeurons, self).__init__(dimensions, dimensions)
 
 
-def split(model, inter_rate, inter_n):  # noqa: C901
+def split(model, loihi_model):  # noqa: C901
```
tbekolay: The splitter accepting `loihi_model` feels like it will make the splitter harder to test once we start to write tests for it. If we're only using the model to get `max_rate` and `dt`, are you ok with modifying this to pass those in instead? Currently also the `dt` is only being used to raise a build error -- this definitely feels like it should be done in the builder and not in the splitter, in which case `split` would only need to accept `max_rate`, which is much easier to test.

nengo_loihi/tests/test_testing.py
```diff
@@ -1,7 +1,32 @@
...
+import numpy as np
 
 
 def test_seed(request, seed):
     if request.config.getoption("seed_offset"):
         pytest.skip()
     assert seed == 2004919645
+
+
+def test_allclose(allclose, capsys):
```
tbekolay: This test makes me realize that it doesn't make sense to compare RMSEs for tests where you're doing `assert not allclose`. I'll make a separate issue for that though.

nengo_loihi/splitter.py
hunse: Ok, fair. In #95 I end up using `inter_tau` here, too, but I guess we can just pass those two (and there might be a way to default that, too, so we don't have to pass it).

PR #56, opened 2018-09-10
Set default precompute to False

### LINKED ISSUE #53: Auto-detect precompute
Right now we have users pass in `precompute=True` or `precompute=False` to change how the model is organized under the hood. There's not much guidance as to why they would choose one vs the other, so it feels odd having this be a user choice.

Ideally, whether the model is to be precomputed or not should be automatically determined, to the point that it is not user configurable.

The next level of this would be to allow the user to specify a preference, but consider that a suggestion and switch when we have reason to believe that a model can be precomputed (or not).

The smallest change to the current situation that would be helpful would be to detect the situation that results in the `Cannot precompute input, as it is dependent on output` exception and automatically switch to `precompute=False` in that situation.

tbekolay:
precomputing node outputs is an optimization which we should not enable by default, as some models cannot be precomputed. Some models will now be slower as a result, but they can be sped back up
by explicitly passing `precompute=True`. This was discussed in #53, but this PR does not close #53.

- [x] Go through test suite and debug any new failures
- [na] Fix test_input_node on hardware


drasmuss:
With `precompute=False` there are four unit tests that hang 100% of the time.  I think that's significant enough that we need to solve that hanging issue before changing the default.  At least with `precompute=True` there is a useful error message if the model is not precomputable.  If we switch to `precompute=False` and people's models just start hanging indefinitely, with no indication what the problem is, that seems like a worse user experience for sure.

tcstewar:
> With `precompute=False` there are four unit tests that hang 100% of the time.

Which unit tests are these?

drasmuss:
They're marked in the latest commit (https://github.com/nengo/nengo-loihi/pull/56/commits/6ecae38ac28eac8fcf810bdcd73ff911cdd0eaf1)

tcstewar:
Just as a reminder, those hanging tests are fixed in https://github.com/nengo/nengo-loihi/pull/78 .  Also, it would be extremely good to have this merged before the INRC workshop, as it greatly simplifies using `nengo_gui` for the tutorials.

tcstewar:
> Waiting on your reply on that PR.

That PR's now been merged...  :)

drasmuss:
Rebased and reran the unit tests, and everything is passing now without hangs (other than those 300-neuron tests that were hanging previously).

tbekolay:
OK, I've rebased this and ran all the tests on the emulator and hardware and they all pass for me now. The RMSE is basically the same on emulator and hardware as well, which is heartening given the recent NxSDK 0.7 upgrade.

I'm :+1: on this PR and ready to merge, but since it's my own PR I can't review it. I'll merge once someone gives this a once over and approves it.

tbekolay:
Oh, and rather than look into `test_input_node` here I made a separate issue to track it so we don't forget.

tbekolay:
Just consolidation.

Issue #59, opened 2018-09-10
vth is too large for some models


tcstewar:
When running some of the benchmark models, I'm getting this error fairly frequently:

```
File "/home/terry/nengo-loihi/nengo_loihi/loihi_api.py", line 140, in validate
    core.validate()
  File "/home/terry/nengo-loihi/nengo_loihi/loihi_api.py", line 193, in validate
    vthProfile.validate(core=self)
  File "/home/terry/nengo-loihi/nengo_loihi/loihi_api.py", line 332, in validate
    assert 0 < self.vth <= VTH_MAN_MAX
```

It seems to be dependent on the random number seed -- adjusting the seed can make the issue go away.   The simplest model I've been able to build which replicates this is here:

```python
import nengo
import nengo.spa as spa
import nengo_loihi
import numpy as np

np.random.seed(2)

model = spa.SPA()
with model:
    n = nengo.Node(None, size_in=1)

    vocab = spa.Vocabulary(1)
    vocab.parse('A+B')
    model.cleanup = spa.AssociativeMemory(vocab)

    nengo.Connection(n, model.cleanup.input)
    probe = nengo.Probe(model.cleanup.output, synapse=0)

with nengo_loihi.Simulator(model, precompute=False) as sim:
    sim.run(0.1)
```

Interestingly, if I do any of the following, I can't get it to fail (after trying 10 different random number seeds):
 - getting rid of the connection from the node to the cleanup
 - reducing the parse to ('A') (i.e. only cleaning up to one item)
 - changing the Probe synapse to None
 - switching to precompute=True

So I'm not sure what's happening here, but I'm hoping someone will have some insights...  :)

hunse:
Ok! I was curious why this only happens on the chip, since I'm pretty sure I checked this earlier in the build process.

Turns out, the problem is not that the `vth` is too large; it's that it's too small! It's the `< 0` part of the assertion that's tripping. There's a similar check in `vth_to_manexp`, but it only checks that `vth >= 0`, so it wasn't catching this.

I'm looking more into why `vth` ends up being zero, but the first thing I've noticed is that this model has some very large weights (the max weight is around `1e17`). This seems to be throwing `discretize` for a loop. Any ideas why the weights might be so large, @tcstewar? I feel like we're going to have trouble representing such large weights no matter what we do in `discretize`.

tcstewar:
[...]

PR #61, opened 2018-09-11
loihi_api: Split validation errors for clarity

hunse:
It will now be clear if a value (e.g. `vth`) is too small
or too large. Before, we lumped these two cases together.
This will help avoid some confusion, like that in #59.

`vth_to_manexp` now checks that `vth > 0` instead of `>= 0`.
This matches the validation done when using Loihi.

### DIFF
--- a/nengo_loihi/loihi_api.py
+++ b/nengo_loihi/loihi_api.py
@@ -28,7 +28,8 @@ V_MIN = -2**23
 def vth_to_manexp(vth):
     exp = VTH_EXP * np.ones(vth.shape, dtype=np.int32)
     man = np.round(vth / 2**exp).astype(np.int32)
-    assert ((man >= 0) & (man <= VTH_MAN_MAX)).all()
+    assert (man > 0).all()
+    assert (man <= VTH_MAN_MAX).all()
     return man, exp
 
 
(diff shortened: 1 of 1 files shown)


nengo_loihi/loihi_api.py
```diff
@@ -309,10 +311,13 @@ def __init__(self, decayV, decayU, refractDelay, enableNoise):
...
-        assert 0 <= self.decayV <= self.DECAY_V_MAX
-        assert 1 <= self.refractDelay <= self.REFRACT_DELAY_MAX
-        assert 0 <= self.enableNoise <= 1
+        assert self.decayU >= 0
+        assert self.decayU <= self.DECAY_U_MAX
+        assert self.decayV >= 0
+        assert self.decayV <= self.DECAY_V_MAX
+        assert self.refractDelay >= 1
+        assert self.refractDelay <= self.REFRACT_DELAY_MAX
+        assert self.enableNoise in (0, 1)
```
drasmuss: Out of curiosity, why do we have to use 0/1 for this instead of True/False?  Something to do with the NxSDK interface?

hunse: Yeah, their interface uses 0/1, so I just used the same. We could use True/False in ours and then convert, since that would be clearer.

Issue #60, opened 2018-09-10
Ensembles with n_neurons>1024 give obscure error message


tcstewar:
Here's the error message you get if n_neurons>1024:

```python
model = nengo.Network(seed=0)
with model:
    state = nengo.Ensemble(1025, 1)
with nengo_loihi.Simulator(model, precompute=True) as sim:
    sim.run(0.01)
```

```
  File "lor.py", line 8, in <module>
    with nengo_loihi.Simulator(model, precompute=True) as sim:
  File "/home/terry/nengo-loihi/nengo_loihi/simulator.py", line 250, in __init__
    self.loihi = self.model.get_loihi(seed=seed)
  File "/home/terry/nengo-loihi/nengo_loihi/loihi_cx.py", line 460, in get_loihi
    return LoihiSimulator(self, seed=seed)
  File "/home/terry/nengo-loihi/nengo_loihi/loihi_interface.py", line 400, in __init__
    self.build(cx_model, seed=seed)
  File "/home/terry/nengo-loihi/nengo_loihi/loihi_interface.py", line 418, in build
    self.n2board = build_board(self.board)
  File "/home/terry/nengo-loihi/nengo_loihi/loihi_interface.py", line 44, in build_board
    build_chip(n2chip, chip)
  File "/home/terry/nengo-loihi/nengo_loihi/loihi_interface.py", line 53, in build_chip
    build_core(n2core, core)
  File "/home/terry/nengo-loihi/nengo_loihi/loihi_interface.py", line 211, in build_core
    build_group(n2core, core, group, cx_idxs, ax_range)
  File "/home/terry/nengo-loihi/nengo_loihi/loihi_interface.py", line 239, in build_group
    n2core.cxCfg[ii].configure(
  File "/home/terry/NxSDK/nxsdk/graph/nodeset.py", line 123, in __getitem__
    return self.nodeInterfaceClassWrapper(self.data[key], self, key)
IndexError: invalid index
```

It'd be nice to give a clearer message (and would be even nicer to automatically split the ensemble across cores, but that's a whole separate issue).  :)

tbekolay:
A clearer error message was added in #62.

PR #62, opened 2018-09-11
Check group size in one-to-one-allocator

hunse:
Addresses #60.

### DIFF
--- a/nengo_loihi/allocators.py
+++ b/nengo_loihi/allocators.py
@@ -1,3 +1,4 @@
+from nengo.exceptions import ValidationError
 import numpy as np
 
 from nengo_loihi.loihi_api import (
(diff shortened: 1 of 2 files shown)


hunse:
The emulator is designed to be agnostic to how the model is actually allocated on the board. So when running on the emulator, the allocation never happens and so you're right, this error doesn't come up. And I'm not sure where we would enforce it for the emulator, since once we have better allocators we will split large CxGroups across multiple cores. (Though until that happens, I guess we could do a check in CxGroup, too.)

I think we might want to have some manual tests of the allocator. These could either be done by creating a `CxModel` manually and passing that to the allocator, or by making a Nengo model and passing it through the builder. We'd then take the output and run some static checks on it. This seems to be the kind of thing that we might want to check in that way.

PR #64, opened 2018-09-12
Remove passthrough nodes onchip

tcstewar:
This PR gets rid of the passthrough Nodes between on-chip Ensembles.  They are replaced by Connections that do the same thing as those Nodes.  Without this PR, all passthrough Nodes are implemented on the host, which means a lot more communication between the host and the chip.  Also, removing these passthrough Nodes allows us to use precompute=True for networks with these intermediate passthrough Nodes.

The basic functionality works, but the main thing I still need to do here is write a *lot* of tests.  In particular, this is a completely different method for removing passthrough nodes than I've ever implemented before.  This method is a lot cleaner, I think, and I think it's much less prone to the weird edge-case problems that I've run into with previous implementations, but given that I'm used to seeing lots of weird edge cases, I want to make sure there's lots of tests to make sure it handles different network configurations and all slicing possibilities and whatnot.

I would also like to do some exploration of what drawbacks moving these passthrough nodes on-chip has.  In particular, I think there are cases where this greatly increases the number of connection weights that must be stored.  I'm not quite sure what these cases are, but we'll see what things come up for the benchmarks.

 - [x] finish unit tests
 - [x] add option to turn this on or off

### DIFF
--- a/nengo_loihi/splitter.py
+++ b/nengo_loihi/splitter.py
@@ -10,15 +10,16 @@ from nengo_loihi.loihi_cx import (
     ChipReceiveNode, ChipReceiveNeurons, HostSendNode, HostReceiveNode,
     PESModulatoryTarget)
 from nengo_loihi.neurons import NIF
+from nengo_loihi.passthrough import convert_passthroughs
 
 logger = logging.getLogger(__name__)
 
 
 def base_obj(obj):
+    if isinstance(obj, nengo.base.ObjView):
+        obj = obj.obj
     if isinstance(obj, nengo.ensemble.Neurons):
         return obj.ensemble
-    elif isinstance(obj, nengo.base.ObjView):
-        return obj.obj
     elif isinstance(obj, nengo.connection.LearningRule):
         return obj.connection
     return obj
@@ -100,9 +101,13 @@ class SplitNetworks(object):
         obj = base_obj(obj)
         logger.debug("Removing %s", obj)
         self.removes.append(obj)
+        if obj in self.adds:
+            del self.adds[obj]
+        elif obj in self.moves:
+            del self.moves[obj]
 
 
-def split(net, precompute, max_rate, inter_tau):
+def split(net, precompute, max_rate, inter_tau, remove_passthrough=False):
     logger.info("Splitting model into host and chip parts")
     networks = SplitNetworks(net, max_rate=max_rate, inter_tau=inter_tau)
 
@@ -110,13 +115,19 @@ def split(net, precompute, max_rate, inter_tau):
     place_nodes(networks)
     place_ensembles(networks)
 
+    # --- Step 1b: remove passthrough nodes
+    if remove_passthrough:
+        conns = merge_passthrough_nodes(networks)
+    else:
+        conns = networks.original.all_connections
+
     # --- Step 2: place simple connections
-    place_internetwork_connections(networks)
+    place_internetwork_connections(networks, conns)
 
     # --- Step 3: split complex connections
-    split_host_to_chip_connections(networks)
-    split_chip_to_host_connections(networks)
-    split_host_to_learning_rules(networks)
+    split_host_to_chip_connections(networks, conns)
+    split_chip_to_host_connections(networks, conns)
+    split_host_to_learning_rules(networks, conns)
 
     # --- Step 4: place precomputable parts of host
     if precompute:
@@ -142,6 +153,8 @@ def place_nodes(networks):
     # Only ChipReceiveNodes can be run on chip
     for node in networks.original.all_nodes:
         if isinstance(node, ChipReceiveNode):
+            # Typically ChipReceiveNodes are created by the splitter, but
+            # it's conceivable that advanced users might make them manually
             networks.move(node, "chip")
         else:
             networks.move(node, "host")
@@ -176,13 +189,37 @@ def place_ensembles(networks):
             networks.move(ens, "chip")
 
 
-def place_internetwork_connections(networks):
+def merge_passthrough_nodes(networks):
+    offchip = set()
+    for obj, target in networks.moves.items():
+        if isinstance(obj, nengo.Node) and obj.output is None:
+            # this is a passthrough Node so don't force it to be offchip
+            continue
+        elif target == 'host':
+            offchip.add(obj)
+
+    remove_nodes, remove_conns, add_conns = convert_passthroughs(
+        networks.original, offchip)
+    for n in remove_nodes:
+        networks.remove(n)
+    for c in remove_conns:
+        networks.remove(c)
+
+    conns = networks.original.all_connections
+    for c in remove_conns:
+        conns.remove(c)
+    conns.extend(add_conns)
+
+    return conns
+
+
+def place_internetwork_connections(networks, conns):
     """Connections from two objects placed in the same location go there.
 
     That is, connections from two objects on the host are done on the host,
     and connections from two objects on the chip are done on the chip.
     """
-    for conn in networks.original.all_connections:
+    for conn in conns:
         pre_loc = networks.location(conn.pre_obj)
         post_loc = networks.location(conn.post_obj)
         if pre_loc == post_loc:
@@ -191,8 +228,8 @@ def place_internetwork_connections(networks):
             networks.move(conn, pre_loc)
 
 
-def split_host_to_chip_connections(networks):
-    for conn in networks.original.all_connections:
+def split_host_to_chip_connections(networks, conns):
+    for conn in conns:
         if conn in networks:
             # Already processed
             continue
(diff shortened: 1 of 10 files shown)


tcstewar:
I've added the basic set of tests I can think of for this.  I still feel like there should be a lot more, especially exploring some of the more complex slicing that can happen in Connections, but the current set covers a pretty wide range of things.  If anyone thinks of other things that should be added, please do so!

tcstewar:
> Can you push any changes you've made to this branch?

I've added my documentation for `passthrough.py`.  I hope that helps!  :)

tcstewar:
I just rebased this to master, fitting it in with the refactor-splitter changes.  I've run it on the benchmark tasks and it seems to all work fine!  :)

nengo_loihi/splitter.py
```diff
@@ -139,6 +146,8 @@ def place_nodes(networks):
     # Only ChipReceiveNodes can be run on chip
     for node in networks.original.all_nodes:
         if isinstance(node, ChipReceiveNode):
+            # TODO: This if branch can never happen as ChipReceiveNodes
```
tbekolay: The reason I put this here is in case we or someone manually makes a `ChipReceiveNode` in the original network. It's exactly the kind of thing that we start off saying "oh, no one would do that." Then later "oh, someone did that, huh." Then even later "oh, I want to do that."

tbekolay:
Generally this is looking good to me, I have a few thoughts that I'll put in to commits shortly.

The main thing that I'm really struggling with is the `remove_passthrough` kwarg to the Simulator. Similar to `precompute`, it just really feels like the kind of thing I don't expect an average user to know what to do with. Especially given that we're defaulting it to `False`, I'm not sure if this is the best way to expose this feature to the user.

First, what's the reason that we're defaulting to `False`? This doesn't feel like a straight-up optimization like the `precompute` flag, this feels more like a change we're making to implement networks better on the chip (we are only removing passthrough nodes on-chip after all). We're already doing a bunch that the user isn't really aware of in the splitter to change the network structure, so my inclination would be to lump removing passthrough nodes in with the rest of the splitter stuff and do it by default.

Second, this is perhaps a bigger conversation for a separate PR, but I'm not convinced that a kwarg to the `Simulator` is the best place to be able to turn on / off this functionality. Simulator kwargs are really front and center; the `Simulator` is the only part of `nengo_loihi` that user **must** interact with, so whatever we expose in that constructor is also implicitly something that we are telling our users that they must understand. I feel like if there is little downside to removing passthrough nodes, we should just do it and if for some debugging reasons we want to disable it, we should do it through modifying something in the splitter (e.g. a class attribute of `SplitNetworks` or something like that). If the concern is that this change isn't a cut-and-dry improvement, then we should figure out why that is.

I'm not sure if these are things that we necessarily want to address now, but would like to hear a bit of the reasoning behind doing it this way to figure out if it's something we want to address now or if we should push those changes to a future release.

tbekolay:
Also can you make sure to run the test suite before handing off PRs? All the tests in `test_passthrough` and a few in `test_splitter` are failing now. It looks like it's because of small API changes and not deep issues, so it'll be quick to fix, but it's a bit disheartening to make some changes and have a bunch of tests fail because they were already failing and not because of the changes I made. Running the whole test suite with the emulator takes 70 seconds on my machine so there's no excuse for not running the tests on a regular basis.

tcstewar:
> Also can you make sure to run the test suite before handing off PRs?

Argh, sorry -- I need to be better about maintaining that habit.  It was late at night, but that's no excuse.  I'll take a stab at fixing it.

tcstewar:
> The main thing that I'm really struggling with is the `remove_passthrough` kwarg to the Simulator. Similar to `precompute`, it just really feels like the kind of thing I don't expect an average user to know what to do with. Especially given that we're defaulting it to `False`, I'm not sure if this is the best way to expose this feature to the user.

I agree.  The main reason that it's an option right now is that it's a big change and I have no idea how it'll affect the other models that are out there.  I'm pretty sure that we'll eventually want to switch it to defaulting to True and remove it from Simulator arguments (so you'd have to do something else to set it to False).  But I'm not sure we want to impose such a big change a few days before a release.

tbekolay:
> But I'm not sure we want to impose such a big change a few days before a release.

That's part of the benefit of having a robust test suite ;) There are always going to be other models out there, but if we're testing as many of them as we can find, then I would prefer to make the plunge and have people file bug reports, which we can then add to the test suite. Having half-tested features that we're not confident about people using is worse than not having that feature at all, IMO.

tcstewar:
> That's part of the benefit of having a robust test suite ;)

Yup!   I guess I was just a bit gun-shy after breaking the keyword-spotting demo.  :)  

> Also can you make sure to run the test suite before handing off PRs?

I just added some changes that fix the test suite.  Sorry again about that!

nengo_loihi/splitter.py
tcstewar: Fair enough.  Although I still think a comment in there to this effect would be helpful -- when I was reading through the code I was confused as I thought this was an expected branch to have happen.  :)

tcstewar: (and I should have caught this during my review of the original PR, but I misinterpretted what was happening then and didn't realize this was happening before ChipReceiveNodes were created)

tbekolay:
I switched the default to be `remove_passthrough=True` and the mean relative RMSE is about the same, which is a good sign!

However, learning is broken because the splitter relies on the `connection.learning_rule` object to match up the learned connection with connections to `connection.learning_rule`, which looks like it changes when passthrough nodes are removed in learning networks.

Should I look into trying to fix this or should we disable removing passthrough nodes if they participate in learning rules in any way?

tcstewar:
> Should I look into trying to fix this or should we disable removing passthrough nodes if they participate in learning rules in any way?

Hmm, interesting....  We should fix this, but I'm not quite sure what the best way is to do it, and I can see a bunch of weird edge cases where we'd have to do things like apply a transform to the learning rule's error signal....  So I'd say for now we don't remove passthroughs where any of the Connections involved have `learning_rule_type!=None`.

tbekolay:
> any of the Connections involved have `learning_rule_type!=None`.

Sounds good. It might be the case that we also have to do this if, for any connection, `isinstance(conn.post_obj, nengo.connection.LearningRule)`, but I'll have to check.

tcstewar:
> Sounds good. It might be the case that we also have to do this if, for any connection, `isinstance(conn.post_obj, nengo.connection.LearningRule)`, but I'll have to check.

Ah, right, because it might be ens->passthroughnode->learning_rule, and we need the ens to be marked as offchip, but right now only the passthroughnode will be automatically marked as offchip, and so when it gets compiled out the ens will not become automatically marked as offchip.....

tbekolay:
Yeah ... I think long term this stuff should be handled under the hood (i.e. in the builder) so for now I think we should just not attempt to handle these cases through frontend manipulations.

tbekolay:
I rebased this to master and pushed a bunch of commits. Summary:

- One of the passthrough tests was failing. This is because the `SplitNetwork.remove` was not expect that something was already added to a network would later be removed, so things were still showing up in the host even though they probably didn't have any connections to them. Modified `remove` to ensure things always get removed.
- I raise an exception when a cluster participates in a learning rule, which allows most of the learning tests to pass... except one. But that one is failing an allclose check, so I'm thinking that removing passthrough nodes changed the seeds of some objects, which makes the behavior randomly worse (but probably not necessarily worse). So, I'll play around with it a little bit to see if it can be fixed without having to widen tolerances.
- The learning rule test that started failing revealed an issue with the `allclose` fixture when comparing two things that require broadcasting (a common e.g. being something like `allclose(bigarray, 0.5)`). Fixed that issue.
- For some reason, this branch also revealed an issue with `test_probedict_fallbacks`, which I fixed. Not sure why that started happening now and not earlier though...

Anyhow, with those changes all tests pass except for that one learning test. There are still some code organization things I want to play around with, but I'm going to leave this branch in this state for now in case other people want to look at it and pick it up again next week.

hunse:
So what's the status on this? Can it be merged? If it's just code organization things that are pending, then I would leave that as future work.

drasmuss:
Is 27bcc2f the first commit in this branch?

drasmuss:
I believe that the test failure is because removing passthrough nodes changes the random seeds for all the other objects in the network. I think this might be be fixed by #70? But it's a non-trivial rebase so I'm still working on that.

I mention it now, because assuming that does solve the problem, then this becomes dependent on #70 and https://github.com/nengo/nengo/pull/1476.  That may be more overhead than we want for the 0.4 release, if we want to get that out tomorrow.

Also, I know I sound like a broken record, but this should hopefully be another issue that is easier to resolve once we move more of the build logic into the backend, as then we will have all the seed information at the same time as we are trying to perform these manipulations.

tbekolay:
> removing passthrough nodes changes the random seeds for all the other objects in the network

That makes sense. The other alternative is to adjust the tolerance, which feels pretty bad but the tolerance should be wide enough that changes in seed don't fail the tests regardless. So I would say that if the only issue is the one test that needs tolerances adjusted then we do that. If there are more issues raised by running the Nengo test suite, then we leave this PR for next release.

drasmuss:
Tests are all passing (or marked as xfail) now, in emulator and hardware, so this is ready for a final review.

Issue #66, opened 2018-09-13
`allclose` should record relative RMSE


hunse:
Right now, our `allclose` fixture records the absolute RMSE between two values. This both makes the RMSE difficult to interpret without knowing the details of the test (i.e. the magnitude of the variables being compared), and it makes the accumulated RMSE across tests much less meaningful.

We should use some form of [relative RMSE](https://en.wikipedia.org/wiki/Root-mean-square_deviation#Normalized_root-mean-square_deviation). Using the range to normalize is sensitive to outliers, and using the mean doesn't make sense for distributions with both positive and negative values. Good options might be to take the mean of the absolute values, or the inter-quartile range.

The other question is how we want to normalize by both inputs `a` and `b` (which we are taking the RMSE between). If we know one (e.g. `b`) is always the "reference" input that we are comparing `a` against, then it might make sense to take the normalizing metric over it. I've always thought that because `np.allclose` is not symmetric (the relative error tolerance is computed using the absolute value of `b`), that we should try to keep `b` as the "reference" input. But I don't think we've really paid much attention to this.

So my inclination would be to normalize by the mean absolute value across both `a` and `b`, i.e. `0.5 * (np.abs(a).mean() + np.abs(b).mean())`.

hunse:
Done in 1e3816f

PR #67, opened 2018-09-13
allclose now reports relative RMSE

hunse:
Addresses #66.

I chose to normalize by the average RMS value of the arguments (makes sense for normalizing RMSE). If one input is all zero, this will always result in a relative RMSE of 1 (unless the other matrix is also zero, in which case it's undefined).


hunse:
Ok, I've rejigged things to avoid the warnings.

hunse:
Note that in #55, I introduce an `assert_allclose` fixture. It will require the same changes as here (ideally, we can rework it to avoid code duplication).

PR #68, opened 2018-09-13
Reset spike probes

drasmuss:
This fixes the spike counter issue pointed out in #5.  Note that there are still some issues with random hanging in the test suite, but I think those are caused by being assigned to particular boards in the INRC cloud, rather than global state (rerunning the tests in the same way but landing on a different device fixes the problem, as opposed to earlier where the tests would deterministically fail at the same spot in the test suite).

I also added a `--no-hang` option to pytest, which will automatically skip tests that are known to hang on Loihi  (`pytest nengo_loihi --target loihi --no-hang`).  That way you can run the whole test suite at once, rather than having to stop and restart.

### DIFF
--- a/nengo_loihi/loihi_interface.py
+++ b/nengo_loihi/loihi_interface.py
@@ -16,12 +16,14 @@ try:
     from nxsdk.arch.n2a.compiler.tracecfggen.tracecfggen import TraceCfgGen
     from nxsdk.arch.n2a.graph.graph import N2Board
     from nxsdk.arch.n2a.graph.inputgen import BasicSpikeGenerator
+    from nxsdk.arch.n2a.graph.probes import N2SpikeProbe
 except ImportError:
     exc_info = sys.exc_info()
 
     def no_nxsdk(*args, **kwargs):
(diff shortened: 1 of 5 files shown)


nengo_loihi/loihi_interface.py
```diff
@@ -397,6 +399,8 @@ def __init__(self, cx_model, seed=None, snip_max_spikes_per_step=50):
         if seed is not None:
             warnings.warn("Seed will be ignored when running on Loihi")
 
+        N2SpikeProbe.probeDict = {}
```
tbekolay: I would lean toward doing `N2SpikeProbe.probeDict.clear()` unless there was a specific reason you went with assigning a new dict; thoughts?

drasmuss: I just wanted to be super sure we were starting from a fresh dictionary that nothing else had a handle to.  But it's also true that if other things _do_ have a handle to the old (broken) probe dict, then `.clear` might fix any issues for them as well.  I'm fine either way :+1:

PR #69, opened 2018-09-18
Limit intercept values

tcstewar:
As part of debugging #59 , @hunse noted that neurons with intercepts near 1 give very high gains, which makes those neurons vastly dominate the dynamic range of the fixed-point representation for the connection weights.  To address this, this PR puts an upper bound on the intercepts.  Any intercept above this limit is reduced to that limit.  In addition, to avoid unnatural clustering of intercepts at that bound, if the intercept distribution is given as a `nengo.dists.Uniform`, and the high value is above the limit, then it is reduced to the limit before sampling.

To illustrate the effects of this fix, here's a simple test model:
```python
model = nengo.Network()
with model:
    stim = nengo.Node(lambda t: np.sin(t*2*np.pi))
    ens = nengo.Ensemble(n_neurons=2000, dimensions=1)
    output = nengo.Node(None, size_in=1)
    nengo.Connection(stim, ens, synapse=None)
    def function(x):
        return x**2
    nengo.Connection(ens, output, function=function, synapse=None)
    p_stim = nengo.Probe(stim)
    p_output = nengo.Probe(output)

sim = nengo_loihi.Simulator(model, intercept_limit=1.0)
sim.run(1.0)

actual = sim.data[p_output]
ideal = function(sim.data[p_stim])
syn = nengo.synapses.Lowpass(0.1)
import pylab
pylab.plot(sim.trange(), syn.filt(actual))
pylab.plot(sim.trange(), syn.filt(ideal))
```

With `intercept_limit=1.0` (which is the same as without this PR), we get this:
image

With `intercept_limit=0.95`, we get this:
image

That's a pretty startling improvement.  :)

To find a good default `intercept_limit`, I ran this a bunch of times and varied the number of neurons:
image

And I varied the order of the polynomial ($x^1$, $x^2$, $x^3$)
image

So, given this, I've set the default to 0.95.

### DIFF
--- a/nengo_loihi/builder.py
+++ b/nengo_loihi/builder.py
@@ -104,6 +104,9 @@ class Model(CxModel):
         self.builder = Builder() if builder is None else builder
         self.build_callback = None
 
+        # limit for clipping intercepts, to avoid neurons with high gains
+        self.intercept_limit = 0.95
+
     def __str__(self):
         return "Model: %s" % self.label
 
@@ -215,7 +218,7 @@ def gen_eval_points(ens, eval_points, rng, scale_eval_points=True):
     return eval_points
 
 
-def get_gain_bias(ens, rng=np.random):
+def get_gain_bias(ens, rng=np.random, intercept_limit=1.0):
     if ens.gain is not None and ens.bias is not None:
         gain = get_samples(ens.gain, ens.n_neurons, rng=rng)
         bias = get_samples(ens.bias, ens.n_neurons, rng=rng)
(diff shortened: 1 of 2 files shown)


drasmuss:
Does the appropriate intercept limit interact with the radius or max rate values?

tcstewar:
> Does the appropriate intercept limit interact with the radius or max rate values?

Hmm, good question....  My expectation is that it shouldn't interact with the radius at all, since the intercept is pre-radius.  That said, I wouldn't be surprised if, for this particular case, adjusting the radius does make the function harder to compute, but that's more an issue with the function, rather than the radius itself....  And max_rate should only linearly affect the gain, so I wouldn't expect anything big there either.... 

But, that's all just my intuition, which has been very wrong before.  :)  I'll run some parameter sweeps and check.

hunse:
I like having the `intercept_limit` configurable in `Model`. I've done the same with the interneuron parameters here: e2ede824b7cd83ae4fae155d98cd0a35ed199110

I agree that we should take it out of `Simulator` arguments. If it's configurable on `Model`, it's easy for a power user to configure ([see here](https://github.com/nengo/nengo-loihi/commit/8f7c7560811928f6946af8e30945fc4588dc9a8b#diff-11c45e7ccbf437c0d0a3f0a0fcedf424R50)).

tbekolay:
Added two commits, one with some minor changes (23c0b1d47b5ff9a6c6db9136bad337907f381768) and one with a test (b2f2fc12e8a7ab7a3dd45f50d1ae7fcbcc817db3). Will merge on @tcstewar's OK, or mid-day tomorrow if there are no objections.

tcstewar:
Awesome!  I like the move to Model-only, and those are some nice improvements to the code.  Thank you!

PR #70, opened 2018-09-18
Seeds match Nengo

hunse:
This ensures that all object seeds are the same as for the corresponding reference Nengo model. It also ensures that seeds for all objects are set and the same whether precompute is True or False.

Depends on nengo/nengo#1476.

Computing the seeds before building, as I'm doing here, is something we've talked about adding to Nengo for a while. I would be open/happy to take the functionality here and move it to Nengo instead, since I think that is ultimately where we want it.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -29,6 +29,13 @@ Release history
   this transition. The MNIST convnet example demonstrates the new syntax.
   (`#142 <https://github.com/nengo/nengo-loihi/pull/142>`__)
 
+**Fixed**
+
+- Objects in nengo-loihi will have the same random seeds as in
+  nengo core (and therefore any randomly generated parameters, such as
+  ensemble encoders, will be generated in the same way).
+  (`#70 <https://github.com/nengo/nengo-loihi/pull/70>`_)
+
 0.5.0 (February 12, 2019)
 =========================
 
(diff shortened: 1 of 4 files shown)


hunse:
This branch is off of #71, but the functionality doesn't depend on it. It was just in trying to get that stuff working that I stumbled upon this problem, so I was using it as a test case. So just look at the last commit when reviewing this.

hunse:
Rebased this to #142, so that it can work with current Nengo.

hunse:
This requires nengo/nengo#1501 to work.

nengo_loihi/tests/test_learning.py
```diff
@@ -208,7 +208,7 @@ def test_multiple_pes(init_function, allclose, plt, seed, Simulator):
 
         probe = nengo.Probe(output, synapse=0.1)
 
-    simtime = 0.5
+    simtime = 0.6
```
hunse: Why?

hunse:
First fixup looks good to me. I added some comments just to make it clear that there's two sections to the test.

I'm curious why the second fixup is necessary. If it's because that test was failing, it just fails randomly. It's fixed in #140.

nengo_loihi/tests/test_learning.py
drasmuss: This test was failing, but I'm pretty sure it was just because the seed changed (running it various times with different seeds it would sometimes randomly fail).  Increasing the time made it pass reliably.

hunse: If switching to 0.6 makes it work for a wider range of seeds, probably worth the change, then. Maybe we can add a short statement in the commit message saying this?

As I mentioned below, the randomness of this test is also fixed in #140, so between the two changes, should be solid!

PR #71, opened 2018-09-18
Neuron->Ensemble connections

hunse:
This allows connections from `Ensemble.Neurons` to an `Ensemble` (and going through the encoders). This was kind of an edge case that we'd forgotten about. I'm not sure if I'll end up needing it for what I'm doing, so maybe we'd rather forbid this case explicitly, since it makes the connection builder even more complicated again (though I've done some stuff in there to clean it up, too).


arvoelke:
To simplify the implementation, can the builder reduce this to the case of `Ensemble -> Ensemble` with `solver = nengo.solvers.NoSolver(transform.T, weights=False)`? Then as long as the `NoSolver` works correctly, this should work as well, unless I'm missing something extra that `Neurons -> Ensemble` enables. I've posted an example in #152.

tbekolay:
> can the builder reduce this to the case of `Ensemble -> Ensemble` with `solver = nengo.solvers.NoSolver(transform.T, weights=False)`

This is what we did in #85 but we had to revert that in #103 because it caused the keyword spotting network to fail. I believe we tracked down the issue to being the magnitude of the weight matrix because it started working again when we scaled it down significantly, but we ran out of time to figure out where in the builder things were being scaled incorrectly and to make sure fixing that scaling didn't break other things.

Issue #72, opened 2018-09-19
Set default solver to use weights


tbekolay:
Wanted to note this in case I forget! We should make this quick one-line change to `nengo_loihi.config.set_defaults`; it might not affect everything, but it will probably improve accuracy on our test suite.

hunse:
I would like to try this on some bigger models before we make the change. It works well for smaller ensembles, but my guess is if we wanted to connect e.g. a 400 neuron population to another 400 neuron population, we'd run out of synapse memory (if I'm calculating correctly, we have `2**14 * 64` bits of memory, and if we're using 8 bits per weight that's about 130000 weights).

Maybe we could have an option that we pass to `set_defaults` that can easily turn this on or off. We could also have a "smart" system that looks at connection sizes and only does this for smaller connections, but that starts to get kinda complicated.

tcstewar:
Hmm, I'm also not quite sure we could do it with a one-line change to `set_defaults`....  setting `nengo.solvers.LstsqL2.weights.default = True` doesn't work:

```python
import nengo
model = nengo.Network()
nengo.solvers.LstsqL2.weights.default = True
with model:
    stim = nengo.Node([0])
    a = nengo.Ensemble(n_neurons=50, dimensions=1)
    b = nengo.Ensemble(n_neurons=50, dimensions=1)
    c1 = nengo.Connection(stim, a)
    c2 = nengo.Connection(a, b)
print(c1.solver)
print(c2.solver)
```
gives you this:
```
LstsqL2(reg=0.1, solver=Cholesky(transpose=None), weights=False)
LstsqL2(reg=0.1, solver=Cholesky(transpose=None), weights=False)
```

Also, doing it with the config system directly reveals this problem:
```python
import nengo
model = nengo.Network()
model.config[nengo.Connection].solver = nengo.solvers.LstsqL2(weights=True)
with model:
    stim = nengo.Node([0])
    a = nengo.Ensemble(n_neurons=50, dimensions=1)
    c1 = nengo.Connection(stim, a)
```
```
Traceback (most recent call last):
  File "c:\users\terry\documents\github\nengo_gui\nengo_gui\page.py", line 224, in execute
    exec(compiled, code_locals)
  File "<nengo_gui_compiled>", line 7, in <module>
  File "c:\users\terry\documents\github\nengo\nengo\base.py", line 35, in __call__
    inst.__init__(*args, **kwargs)
[...]

PR #74, opened 2018-09-19
Force Loihi solvers to weights=True

tcstewar:
This is an initial attempt at implementing a way to force nengo-loihi to do full connection weight matrices rather than whatever the solver says.

I'm posting this not as a polished PR, but rather to get feedback on whether this seems to be the right way to go about implementing this.

As discussed in #72, this cannot be done by adjusting nengo defaults, since nengo currently does not allow you to set a solver with weights=True when you have a Node involved in the Connection.  Furthermore, even if you could do that, with the advent of #64 , we'll be removing the passthrough nodes from models, so the Ensemble->Ensemble connection that we want to have weights=True for won't even exist at that time.  For these reasons, I've implemented this forcing of using weights=True as part of the builder itself.

TODO:
 - [ ] Discuss whether this is the best approach
 - [ ] Write some unit tests
 - [ ] Analyze the effects on accuracy and memory usage for a few benchmark models


tcstewar:
(note: this is branched off of `limit-intercepts`, so just look at the most recent commit for the changes done here)

drasmuss:
Xuan seems to be getting worse behaviour with `weights=True` in #90, so we should figure out why that is before we force `weights=True`.

hunse:
This was more of an issue before we improved the DecodeNeurons (#132).

tbekolay:
As @hunse says, this is less of an issue now, and this part of the code has changed a bunch so rebasing will be difficult. It's still possible to do this across the model by changing the default solver to a `weights=True` solver, so I'm going to close this.

arvoelke:
> As @hunse says, this is less of an issue now, and this part of the code has changed a bunch so rebasing will be difficult.

Do we have anything in place (or an open issue) to test the difference between `weights=True` and `weights=False` to help spot a regression, and as a basis for future improvements?

> It's still possible to do this across the model by changing the default solver to a `weights=True` solver, so I'm going to close this.

As @tcstewar said in the original post this doesn't work for any model which contains a node (which is basically every model that we have). e.g.,

```python
import nengo
    model.config[nengo.Connection].solver = nengo.solvers.LstsqL2(weights=True)
    nengo.Connection(nengo.Node(0), nengo.Ensemble(1, 1))
```

```
nengo.exceptions.ValidationError: Connection.solver: weight solvers only work for connections from ensembles (got 'Node')
```

tbekolay:
> Do we have anything in place (or an open issue) to test the difference between `weights=True` and `weights=False` to help spot a regression, and as a basis for future improvements?

I don't believe that we do, but also I didn't search for it. Feel free to make a new issue if it hasn't been tracked though -- we're trying to keep our PR queue to a minimum so we would prefer new issues over keeping PRs open as a reminder of some issue.

> this doesn't work for any model which contains a node

Good point. You can still change it on a per-connection basis, but you can't set it as a default.

Issue #77, opened 2018-09-21
Spike probes with precompute=True hang for neurons other than LIF(tau_ref=0.002)


tcstewar:
This is the cause of the hanging tests identified in 6ecae38ac28eac8fcf810bdcd73ff911cdd0eaf1 and it is due to this old issue #6 

I'm not sure why it just hangs rather than giving the exception and/or assertion failure that should be happening.

To be more specific, this is the problematic code:
```python
                       if isinstance(probe.target, nengo.ensemble.Neurons):
                            t_ref = probe.target.ensemble.neuron_type.tau_ref
                            assert t_ref == 0.002
                            # TODO: generalize this to other tau_ref values
                            x = (x == 384)
```

That will give an assertion failure for LIF with tau_ref != 0.002, and it'll give an Exception for neuron types that don't have tau_ref (i.e. SpikingRectifiedLinear)

hunse:
Addressed by #78.

PR #78, opened 2018-09-21
Handle different refractory period in precompute=False spike probes

tcstewar:
This fixes the 4 hanging tests @drasmuss identified in https://github.com/nengo/nengo-loihi/commit/6ecae38ac28eac8fcf810bdcd73ff911cdd0eaf1

Those 4 tests now all pass with `precompute=False`

Fixes Issue #77  and Issue #6

### DIFF
--- a/nengo_loihi/simulator.py
+++ b/nengo_loihi/simulator.py
@@ -10,6 +10,7 @@ from nengo.exceptions import ReadonlyError, SimulatorClosed, ValidationError
 from nengo.utils.compat import ResourceWarning
 
 from nengo_loihi.builder import Model, INTER_RATE, INTER_N
+from nengo_loihi.loihi_cx import CxGroup
 import nengo_loihi.config as config
 import nengo_loihi.splitter as splitter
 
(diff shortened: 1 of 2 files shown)


hunse:
I'm wondering if instead of getting the tau_ref from the neuron_type and converting it, we can just grab the refactDelay from the cx_group. This is ultimately what's controlling the refractory period here, so if we use that then we don't have to worry e.g. if we were to change how tau_ref is converted to an integer, or if we allow tau_ref to be a distribution down the road.

tcstewar:
> I'm wondering if instead of getting the tau_ref from the neuron_type and converting it, we can just grab the refactDelay from the cx_group

Ooo, good idea.  Can we get that from the `cx_probe` object?

nengo_loihi/simulator.py
```diff
@@ -602,20 +602,11 @@ def handle_chip2host_communications(self):  # noqa: C901
...
-                            #  where we are in the refractory period
-                            x = (x == (ref_period_in_steps + 1) * 128)
-                        else:
-                            # interneurons have a refractory period of 1 step
-                            x = (x == 128)
+                        # Loihi uses the voltage value to indicate where we
+                        # are in the refractory period. We want to find neurons
+                        # starting their refractory period.
+                        refract_delays = cx_probe.target.refractDelay
+                        x = (x == refract_delays * 128)
```
tcstewar: Nice!  So much cleaner.  Thank you!

hunse: I wasn't quite done. I've added another commit for the case that `cx_probe.target` is not a CxGroup. I'm not sure if this ever actually happens, but it currently is possible to have a CxProbe on a CxSpikeInput (though I think we handle that all on the CPU side, which is probably why it never happens).

hunse:
In the process of trying to test my changes here, I realized that `test_communication.py:test_input_node` is failing on Loihi (at least for me). Has anybody else had this? It didn't seem isolated to this branch; I went back to 0.2 and found it happening.

tbekolay:
Yes, `test_input_node` has been broken for quite some time.

hunse:
Maybe we should mark it as xfail?

tbekolay:
Sure, feel free to mark it xfail for now.

hunse:
Ok, done. It seems to be somewhat random when it fails and passes, so we get some XPASS tests now, but it is less confusing. Also, it seems to all pass when `precompute=False` (it's just much slower), so that's reassuring.

tcstewar:
> but the tests marked as hanging in [e7a18f3](https://github.com/nengo/nengo-loihi/commit/e7a18f35f1038ac77a676baa0376a46a88db0b04) are still reliably hanging for me. `test_voltage_decode` always hangs, and `test_pes_comm_channel` hangs for 300 neurons.
> 
> Do you want me to merge this is as in improvement, or do you want to look into those two tests first @tcstewar?

I just took a quick look at those other two tests, and it looks like they have nothing at all to do with this -- in fact, they're not even related to the `precompute` setting.  But, I have identified a core symptom behind those two other failing tests, so I'll add that as a separate issue and this PR can get merged in as an improvement.  :)  Thanks for pointing out those other tests!

tbekolay:
Awesome, can't believe it was the 300 neurons that is causing the hang :dizzy_face: 

Will merge this in now.

PR #79, opened 2018-09-21
Support Distributions on Connection transforms

### LINKED ISSUE #65: Error when using Distribution for transform
Here's a minimal example:

```
import nengo
import nengo_loihi

with nengo.Network() as net:
    a = nengo.Node([0])
    b = nengo.Ensemble(10, 1)
    nengo.Connection(a, b, transform=nengo.dists.Uniform(-1, 1))

with nengo_loihi.Simulator(net) as sim:
    pass
```

Gives error:
```
Traceback (most recent call last):
  File ".../nengo_loihi/test2.py", line 9, in <module>
    with nengo_loihi.Simulator(net) as sim:
  File "...i\nengo_loihi\simulator.py", line 168, in __init__
    network, INTER_RATE, INTER_N)
  File "...\nengo_loihi\splitter.py", line 208, in split
    transform=c.transform * scaling)
TypeError: unsupported operand type(s) for *: 'Uniform' and 'float'
```

drasmuss:
Previously passing a `Distribution` for a transform gave errors on chip->host and host->chip connections, this fixes that and adds a test.

Note that the way the seeding is done isn't strictly correct, because we don't have access to the connection seeds (since this is all occurring before the seeding happens in the builder).  But it has the same functional behaviour (deterministic output if a seed is set on the Network or Connection).

This fixes #65

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -26,6 +26,7 @@ Release history
 
 - Models can now use the ``nengo.SpikingRectifiedLinear`` neuron model
   on both the emulator and hardware backends.
+- Added support for Distributions on Connection transforms.
 
 **Fixed**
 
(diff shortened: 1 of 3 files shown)


hunse:
#70 might help with the seeding problems.

PR #81, opened 2018-09-21
Convnets everywhere

hunse:
This branch combines work for convnets in the emulator and convnets on the Loihi chip itself. @drasmuss and I are working on separate branches right now so we don't step on each other's toes, but as we accomplish things we'll try to move them into this branch.


hunse:
Ok, so this now has an MNIST network that works well on the emulator. However, it still does not provide the correct output on the chip for unknown reasons.

I'll keep looking into this and see if I can track it down. `test_conv:test_conv_split` appeared to pass the first time I ran it, but now does not. So maybe my answer lies somewhere in there. (I just ran it a few more times, and now it seems to be passing... this isn't good.)

But it could be that people will have to just do deep network stuff in the emulator for now. I'm happy to move this branch somewhere more accessible (we've got many branches that start with `conv2d`) so that people can play around with it, as well as moving the example to a more prominent location. Thoughts?

celiasmith:
Attempting to run the mnist_convnet.ipynb results in the assertion error: 
```~/Documents/nengo/nengo-loihi/nengo_loihi/conv.py in __init__(self, ops, signals, config)
    360             super(Conv2DIncBuilder, self).__init__(ops, signals, config)
    361 
--> 362             assert len(ops) == 1
    363 
    364             self.conv = ops[0].conv2d_transform

AssertionError: 
```
I am on the suggested nengo_dl branch.   This occurs in block 7 of the code.

celiasmith:
Checked the `mnist_convnet.ipynb` example and Dan's updated it to account for the issues i ran into.  So that bit is reviewed.

hunse:
Ok, this is officially ready for full review. All the new tests are passing in emulator and on Loihi (I haven't exhaustively ran the old tests on Loihi, so someone should do that as part of reviewing this).

The way I've organized the commits is that I have all the fixes/improvements that are required for this work at the beginning, then the convnet stuff itself. The convnet stuff really starts around e305d55 when we start fixing up `loihi_cx` to allow population spikes. Then we extend population spikes to Loihi, then we do the Conv2D transform itself along with tests and my original MNIST example. Then we have the example notebook as the final product.

One thing that is still missing is for a number of the tests, I use an MNIST digit to have something realistic to apply convolution to. I created a file with 10 images called `mnist10.pkl`, but I haven't yet added this to the repository. I actually only use the first image, so I'm wondering if I just take that image and put it in a pickle. It would be quite small (the 10 are 31K, so I'd expect around 3K). Would we be okay adding that to the repository?

hunse:
Also, Travis CI is unhappy because I use `nengo_extras` for plotting in some of my conv tests and it doesn't have it. Could we have it for these tests? It makes the plots much nicer.

hunse:
Assigning this to @tcstewar, since Trevor is busy and Dan and I are both authors.

I'm going to go through right now and put the MNIST digit used in the tests in a file by itself and add it to the repo.

hunse:
Rebased this to master and #121 in `conv2d-master-rebase` branch. I haven't tested anything there, yet, though.

hunse:
Ok, this is rebased in `conv2d-master-rebase`. However, it's rebased to #137, since I needed that in there for doing CIFAR-10.

Are we planning on getting #137 into 0.4 (we didn't talk about it). If not, I could get this rebased back onto conv2d-prep like we talked about.

tbekolay:
> Are we planning on getting #137 into 0.4 (we didn't talk about it)

I wasn't planning on it, but let me look at the diff after merging #133.

tbekolay:
> I wasn't planning on it, but let me look at the diff after merging #133.

I looked at it briefly, it's quite a few changes but they look pretty straightforward so I'll give it a closer look tonight.

PR #85, opened 2018-09-25
Move manual decoders onchip

tcstewar:
This PR makes a quick change to the splitter system such that this:

```python
nengo.Connection(a.neurons, output, transform=w)
```
is the same as
```python
nengo.Connection(a, output, solver=nengo.solvers.NoSolver(w.T))
```

The main reason for this is that a) these two should be the same, and b) there's a nice sequence of tutorials @xchoo and I have planned that requires this syntax (we want to teach learning before teaching NEF-style decoders, to match more closely what typical neural network modellers are used to).

### DIFF
--- a/nengo_loihi/splitter.py
+++ b/nengo_loihi/splitter.py
@@ -191,7 +191,7 @@ def split(model, inter_rate, inter_n):  # noqa: C901
                         intercepts=[-1] * dim + [-1] * dim)
 
                     # scale the input spikes based on the radius of the
-                    #  target ensemble
+                    # target ensemble
                     if isinstance(c.post_obj, nengo.Ensemble):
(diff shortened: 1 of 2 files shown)


tcstewar:
> This works for me on the emulator, but on hardware I get:
> 
> ```
> AttributeError: 'PESModulatoryTarget' object has no attribute 'receive'
> ```
> in `handle_host2chip_communications` at the `receiver.receive` line.

It looks like that's because you're calling the Simulator with `precompute=True`, and PES learning has never worked with precompute=True.  Also, since we still don't have multidimensional learning working, we need to keep `post_dims=1`. 

With those two changes, the test passes on both target=loihi and target=sim for me!

PR #87, opened 2018-09-25
Fix number of compartments

hunse:
Two little bugs that I found when finding the 300 neuron hanging bug.

The first is just that we're counting the number of compartments wrong in loihi_interface. We need to add 1 to the last index to get the number.

The second I'm less sure about, but it looked wrong. It's that when we're sending spikes for precompute=True, we gather up all the spikes, and then send them all to the spike generator for the last input. As it happens, this isn't a problem right now, since all inputs share the same spike generator. But if we did change things so that they had their own generators, I think we'd want it as I have it here. I'd like @tcstewar to confirm this, though, since he wrote that code I think.

### DIFF
--- a/nengo_loihi/loihi_interface.py
+++ b/nengo_loihi/loihi_interface.py
@@ -213,7 +213,7 @@ def build_core(n2core, core):  # noqa: C901
 
         for group, cx_idxs, ax_range in core.iterate_groups():
             build_group(n2core, core, group, cx_idxs, ax_range)
-            n_cx = max(max(cx_idxs), n_cx)
+            n_cx = max(max(cx_idxs) + 1, n_cx)
 
     for inp, cx_idxs in core.iterate_inputs():
         build_input(n2core, core, inp, cx_idxs)


tcstewar:
The first fix looks fine to me.  However, the second one breaks every model with more than one input.  The spikes added to a spike generator must be added in temporal order.  That's why the original code collected them all together and sorted them.  Here's a quick example:

```python
model = nengo.Network()
with model:
    stim1 = nengo.Node(1)
    stim2 = nengo.Node(0.5)

    a = nengo.Ensemble(100, 1)
    b = nengo.Ensemble(100, 1)

    nengo.Connection(stim1, a)
    nengo.Connection(stim2, b)

    pa = nengo.Probe(a, synapse=0.1)
    pb = nengo.Probe(b, synapse=0.1)

sim = nengo_loihi.Simulator(model, precompute=True)
with sim:
   sim.run(1)

print(sim.data[pa][-10:])
print(sim.data[pb][-10:])
```

On this branch, the output is:
```
[[0.97023232]
 [0.97219265]
 [0.97007911]
 [0.9701084 ]
 [0.96640179]
 [0.972825  ]
 [0.97377368]
 [0.96872533]
 [0.96706911]
 [0.97483415]]
[[0.00580812]
 [0.00575023]
 [0.00632653]
 [0.00550584]
 [0.00608458]
 [0.00602413]
 [0.00596419]
 [0.00594211]
 [0.00523705]
 [0.0054458 ]]
```

Issue #88, opened 2018-09-26
Emulator crashes when V or U are out of range


tcstewar:
`V` and `U` are both bounded in Loihi, but right now if you go outside of these bounds in the emulator, then an exception is raised. 

For example:
```
model = nengo.Network(seed=3)
with model:
    stim = nengo.Node(1)
    a = nengo.Ensemble(100, 1)
    nengo.Connection(stim, a)
    b = nengo.Ensemble(100, 1)
    nengo.Connection(a, b, transform=10, synapse=0.1)
```

When run, this raises this exception:
```
Traceback (most recent call last):
  File "c:\users\terry\documents\github\nengo-loihi\nengo_loihi\simulator.py", line 409, in run_steps
    self.simulator.run_steps(steps)
  File "c:\users\terry\documents\github\nengo-loihi\nengo_loihi\loihi_cx.py", line 740, in run_steps
    self.step()
  File "c:\users\terry\documents\github\nengo-loihi\nengo_loihi\loihi_cx.py", line 693, in step
    raise SimulationError("Overflow in V (max was %d)" % self.v.max())
nengo.exceptions.SimulationError: Overflow in V (max was 8459846)
```

I believe the solution should be to clip V and U, rather than raising exceptions in `loihi_cx.py`:
```python
        if np.any(u2 > U_MAX):
            raise SimulationError("Overflow in U (max was %d)" % u2.max())
        if np.any(u2 < U_MIN):
            raise SimulationError("Underflow in U (min was %d)" % u2.min())

        # self.V[:] = self.decayV_fn(v, self.decayV, a=12) + u2
        self.v[:] = self.decayV_fn(self.v, u2)
        if np.any(self.v > V_MAX):
            raise SimulationError("Overflow in V (max was %d)" % self.v.max())
        if np.any(self.v < V_MIN):
            raise SimulationError("Underflow in V (min was %d)" % self.v.min())
```

tbekolay:
We added this in 54954fb035c57ecaf48603cbb58738099a0b273b because we were getting very strange behavior on some models on hardware, which turned out to be because some weights were very large and was causing overflows. The explicit exceptions allow us to notice this for debugging early on rather than seeing weird, difficult to debug behavior on hardware.

tbekolay:
I mean, any model that raises this exception is unlikely to run correctly on hardware so it depends on whether that matters for these particular examples... @hunse added the check and exception, any thoughts about other ways we can avoid overflows?

[...]

Issue #83, opened 2018-09-24
Discretize to use as much of the dynamic range as possible


tbekolay:
Right now our discretization process uses the maximum and minimums from the actual values of several quantities, including `intercepts`, which led to issues such as those solved in #69. We should look into other methods for setting the ranges for these discretization processes, as the numerical resolution is not even across the number line, and even if it were we should not be super sensitive to outliers.

PR #89, opened 2018-09-26
Multidimensional learning

tcstewar:
This PR fixes multidimensional learning.  It is based off of `move-manual-decoders-onchip`, but only because that had a nice test that fit for this PR.

All this does is fix the snip code and the communication code such that all the error data gets sent across and applied.  It doesn't change the emulator at all, since it was already working in the emulator.

The communication protocol is not wonderful.  The message format is `[core_id, n_vals, val_0, val_1, ... val_n-1]`, so the first two values are the same every time a message is sent (this isn't much worse than the old format, which was `[core_id, val_0]`).  Optimizing that will be a different PR, and will interact in interesting ways with #26 .  But this is a definite improvement in the meantime.

This is heavily based on, but replaces #22 (the rebasing got too complicated for me so I started this one fresh).

TODO
 - [x] get multiple learning cores working

### DIFF
--- a/nengo_loihi/loihi_interface.py
+++ b/nengo_loihi/loihi_interface.py
@@ -536,9 +536,14 @@ class LoihiSimulator(object):
         # --- generate custom code
         # Determine which cores have learning
         n_errors = 0
+        total_error_len = 0
+        max_error_len = 0
         for core in self.board.chips[0].cores:  # TODO: don't assume 1 chip
             if core.learning_coreid:
+                error_len = core.groups[0].n // 2
+                max_error_len = max(error_len, max_error_len)
                 n_errors += 1
+                total_error_len += 2 + error_len
 
         n_outputs = 1
         probes = []
@@ -564,6 +569,7 @@ class LoihiSimulator(object):
         code = template.render(
             n_outputs=n_outputs,
             n_errors=n_errors,
+            max_error_len=max_error_len,
             cores=cores,
             probes=probes,
         )
@@ -590,7 +596,7 @@ class LoihiSimulator(object):
             phase="preLearnMgmt",
         )
 
-        size = self.snip_max_spikes_per_step * 2 + 1 + n_errors*2
+        size = self.snip_max_spikes_per_step * 2 + 1 + total_error_len
         logger.debug("Creating nengo_io_h2c channel")
         self.nengo_io_h2c = self.n2board.createChannel(b'nengo_io_h2c',
                                                        "int", size)
--- a/nengo_loihi/simulator.py
+++ b/nengo_loihi/simulator.py
@@ -426,9 +426,6 @@ class Simulator(object):
                     self.handle_chip2host_communications()
 
                 logger.info("Waiting for completion")
-                self.loihi.nengo_io_h2c.write(1, [0])
-                self.loihi.nengo_io_h2c.write(1, [0])
-                self.loihi.nengo_io_h2c.write(1, [0])
                 self.loihi.wait_for_completion()
                 logger.info("done")
             else:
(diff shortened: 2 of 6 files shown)


tbekolay:
@tcstewar You removed your assignment, but the PR is still marked as WIP. Is it ready for review? Or collaboration?

tbekolay:
Another thing TODO:

- [ ] Go through the review comments in #22 and ensure they're addressed here.

tcstewar:
> Is it ready for review? Or collaboration?

I'll rebase it to master now that the manual decoders are merged in, then take a pass through the #22 comments, and then it should be ready for review.  :)

tcstewar:
The last commit adds a test of having multiple learning connections at once.  It seems to mostly work, but doesn't converge to the target values and I'm not sure why.  If someone else could take a look at it before the workshop, that'd be great!

tbekolay:
So adding some plotting to that test, I get the following in the emulator

[test_learning.test_multiple_pes_emu.pdf](https://github.com/nengo/nengo-loihi/files/2426387/test_learning.test_multiple_pes_emu.pdf)

and this on hardware

[test_learning.test_multiple_pes.pdf](https://github.com/nengo/nengo-loihi/files/2426388/test_learning.test_multiple_pes.pdf)

One difference is that they change at different rates. That is likely down to a mismatch in learning rates that we probably need to do a better job of matching. It may also be partly due to the fact that error values are clipped to the [-1, 1] range.

The more salient difference is that after a steady, predictable climb toward the target value, the line starts going haywire. This becomes more apparent when you run the network for longer, or with a high learning rate. The fact that the lines all start moving in a predictable manner tells me that it is not likely a problem with the PES implementation (i.e., the delta is being calculated correctly and is being applied to the right synapses, etc etc). I think the problem is most likely some value over or underflowing its discretized range; this has happened before (and is why we introduced the checks discussed in #88). I do not get any warnings raised in the emulator, so the over/underflowed quantity is not U or V; most likely it's the weights themselves. In the emulator, we currently use `float64`s to represent the learning trace and while we initially discretize the weights to match hardware, we don't re-discretize every time the weights change during the simulation, so likely we're using the full `int32` range. Put another way, the problem is that we discretize weights based on their initial value, but when doing online learning, their final values are going to be very different from their initial values.

In the long term, we should definitely think about ways in which we can mitigate this problem. I'll make an issue to start thinking about that.

In the short term, I'm thinking if that the initial function that the learned connection represents results in larger decoders, then we should do a much better job of discretizing. I'll try doing that now.

tbekolay:
Modifying the function (actually in this case just removing the `function=lambda x: 0`) is a success!

Emu: 
[test_learning.test_multiple_pes_emu.pdf](https://github.com/nengo/nengo-loihi/files/2426458/test_learning.test_multiple_pes_emu.pdf)

Hardware: 
[test_learning.test_multiple_pes.pdf](https://github.com/nengo/nengo-loihi/files/2426461/test_learning.test_multiple_pes.pdf)

Also note that they both converge significantly faster. I suspect this might happen on reference Nengo too because the weights might have less distance to travel because they don't start near zero. But, it may also be the case that the range discretization results in the same weight updates having a larger effect than they would have in reference Nengo because weights are being pushed farther as a result of the same delta.

In any case, I'll add some asserts and push the now working test. I'll also make a separate issue to figure out the best way to inform users about this issue.

tbekolay:
OK, pushed 6eab5e3a1c009b5c0f0bc120b364d89047459d7f, and also rebased. If you get a chance, let me know what you think @tcstewar, but in any case I'll do a proper review tomorrow so we can get this merged in.

tcstewar:
> OK, pushed [6eab5e3](https://github.com/nengo/nengo-loihi/commit/6eab5e3a1c009b5c0f0bc120b364d89047459d7f), and also rebased. If you get a chance, let me know what you think @tcstewar, but in any case I'll do a proper review tomorrow so we can get this merged in.

Awesome!  I completely didn't think of that weight scaling issue....  makes a lot of sense in hindsight, but I was not thinking in that direction at all.....   Very good thing to know for the learning tutorials too.  :)

The commit looks good to me. :)  Thank you!

hunse:
It would be nice if we could replicate that chip learning behaviour in the simulator, before we got rid of `function=lambda x: 0`. I assume it's some kind of clipping that's going on. Doesn't have to be for this PR, but if not now then let's make an issue.

tbekolay:
Yeah I'll make an issue.

tbekolay:
Added comments to #83 and #98 and made #99 to track things that we should do as a result of the issues raised in this PR. Squashing and merging now!

PR #92, opened 2018-09-26
Compatibility with NxSDK 0.7 release

drasmuss:
Update `async` parameter to `aSync`. I left our parameter name as `async`.  We could update it, but since we're sticking to the Python convention of snake_case names I figured `async` was preferred.

Updated the installation instructions for the new workflow (and using conda for everything).  This removed the distinction between the INRC/Superhost installation steps, so I combined them.

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -12,10 +12,7 @@ that has ``pip``.
 .. code-block:: bash
 
    git clone https://github.com/nengo/nengo-loihi.git
-   cd nengo-loihi
-   pip install -e .
-
-Note that the ``.`` at the end of ``pip`` command is required.
+   pip install -e nengo-loihi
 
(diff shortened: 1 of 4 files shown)


tbekolay:
Also, since we're now requiring NxSDK 0.7 we should add a check somewhere (probably `LoihiSimulator` so it doesn't get raised for the emulator) and raise an error if we're using an earlier version (and maybe warn for future versions?)

drasmuss:
I moved all of the conftest material to the root in #56, so that we don't have to keep worrying about where we should put things.  So we could just let that happen there, but no harm doing it here either 🤷‍♂️ .

Everything else looks good!

tbekolay:
Oh nice, I'll revert that and look at that PR next.

PR #95, opened 2018-09-27
 Do Node filtering before interneurons 

### LINKED ISSUE #94: Cannot change tau_s on already configured neurons (tau != 5ms)
I am trying to run a pretty standard network that works with the reference simulator, but gives this confusing error with the loihi emulator:

```
/home/arvoelke/CTN/nengo-loihi/nengo_loihi/builder.py:239: UserWarning: Intercepts are larger than intercept limit (0.95). High intercept values cause issues when discretizing the model for running on Loihi.
  "the model for running on Loihi." % intercept_limit)

---------------------------------------------------------------------------
BuildError                                Traceback (most recent call last)
<ipython-input-7-9d2e83577abb> in <module>()
----> 1 with loihi.Simulator(model, dt=dt, precompute=True) as sim:
      2     sim.run(length*dt)

~/CTN/nengo-loihi/nengo_loihi/simulator.py in __init__(self, network, dt, seed, model, precompute, target)
    181 
    182             # Build the network into the model
--> 183             self.model.build(network)
    184 
    185         self._probe_outputs = self.model.params

~/CTN/nengo-loihi/nengo_loihi/builder.py in build(self, obj, *args, **kwargs)
    112 
    113     def build(self, obj, *args, **kwargs):
--> 114         built = self.builder.build(self, obj, *args, **kwargs)
    115         if self.build_callback is not None:
    116             self.build_callback(obj)

~/CTN/nengo-loihi/nengo_loihi/builder.py in build(cls, model, obj, *args, **kwargs)
    139                 "Cannot build object of type %r" % type(obj).__name__)
    140 
--> 141         return cls.builders[obj_cls](model, obj, *args, **kwargs)
    142 
    143     @classmethod

~/CTN/nengo-loihi/nengo_loihi/builder.py in build_network(model, network)
    189     logger.debug("Network step 3: Building connections")
    190     for conn in network.connections:
--> 191         model.build(conn)
    192 
    193     logger.debug("Network step 4: Building probes")

~/CTN/nengo-loihi/nengo_loihi/builder.py in build(self, obj, *args, **kwargs)
    112 
    113     def build(self, obj, *args, **kwargs):
--> 114         built = self.builder.build(self, obj, *args, **kwargs)
    115         if self.build_callback is not None:
    116             self.build_callback(obj)

~/CTN/nengo-loihi/nengo_loihi/builder.py in build(cls, model, obj, *args, **kwargs)
    139                 "Cannot build object of type %r" % type(obj).__name__)
    140 
--> 141         return cls.builders[obj_cls](model, obj, *args, **kwargs)
    142 
    143     @classmethod

~/CTN/nengo-loihi/nengo_loihi/builder.py in build_connection(model, conn)
    703         model.objs[conn]['mid_axons'] = mid_ax
    704 
--> 705         post_cx.configure_filter(post_tau, dt=model.dt)
    706     elif isinstance(conn.post_obj, Node):
    707         raise NotImplementedError()

~/CTN/nengo-loihi/nengo_loihi/loihi_cx.py in configure_filter(self, tau_s, dt, default)
    182         if self.decayU_set and not np.allclose(decayU, self.decayU):
    183             raise BuildError(
--> 184                 "Cannot change tau_s on already configured neurons")
    185 
    186         self.decayU[:] = decayU

BuildError: Cannot change tau_s on already configured neurons
```

Unsure where this is coming from. I'm on the current `master` branch, and here's the code:

```python
import matplotlib.pyplot as plt
import seaborn as sns

import numpy as np

import nengo
import nengo_loihi as loihi
from nengolib import Lowpass
from nengolib.networks import readout
from nengolib.signal import Balanced
from nengolib.synapses import PadeDelay, pade_delay_error, ss2sim

theta = 0.1
order = 3
freq = 3
power = 1.5

print("PadeDelay(%s, %s) => %f%% error @ %sHz" % (
    theta, order, 100*abs(pade_delay_error(theta*freq, order=order)), freq))
pd = PadeDelay(theta=theta, order=order)

# Heuristic for normalizing state so that each dimension is ~[-1, +1]
rz = Balanced()(pd, radii=1./(np.arange(len(pd))+1))
sys = rz.realization

# Compute matrix to transform from state (x) -> sampled window (u)
t_samples = 100
C = np.asarray([readout(len(pd), r)
                for r in np.linspace(0, 1, t_samples)]).dot(rz.T)
assert C.shape == (t_samples, len(sys))

length = 1000
dt = 0.001

process = nengo.processes.WhiteSignal(
    period=length*dt, rms=power, high=freq, y0=0, seed=0)
t = process.ntrange(length, dt=dt)
u = process.run_steps(length, dt=dt)
x = sys.X.filt(u, dt=dt)  # discretizes sys using ZOH

def plot(t, u, x, C=C):
    print("Radii:", np.max(np.abs(x), axis=0))
    w = C.dot(x.T)

    fig, ax = plt.subplots(2, 1, sharex=True, figsize=(14, 8))
    for c, w_i in zip(sns.color_palette('GnBu_d', len(C)), w[::-1]):
        ax[0].plot(t, w_i, c=c, alpha=0.7)
    ax[0].plot(t, u, c='green', linestyle='--', lw=3)
    ax[1].plot(t, x)
    ax[-1].set_xlabel("Time (s)")
    fig.show()

plot(t, u, x)

n_neurons = 128  # per dimension
tau = 0.018329807108324356

map_hw = ss2sim(sys, synapse=Lowpass(tau), dt=None)
assert np.allclose(map_hw.A, tau*sys.A + np.eye(len(sys)))
assert np.allclose(map_hw.B, tau*sys.B)

    u = nengo.Node(output=process, label='u')
    p_u = nengo.Probe(u, synapse=None)

    Bu = [nengo.Node(output=lambda _, u, b_i=map_hw.B[i].squeeze(): b_i*u,
                     size_in=1, label='Bu[%d]' % i)
          for i in range(len(sys))]
    
    X = []
    for i in range(len(sys)):
        X.append(nengo.Ensemble(
            n_neurons=n_neurons, dimensions=1, label='X[%d]' % i))
        
    P = []
    for i in range(len(sys)):
        nengo.Connection(u, Bu[i], synapse=None)
        nengo.Connection(Bu[i], X[i], synapse=tau)
        for j in range(len(sys)):
            nengo.Connection(X[j], X[i], synapse=tau,
                             function=lambda x_j, a_ij=map_hw.A[i, j]: a_ij*x_j)
        P.append(nengo.Probe(X[i], synapse=None))

with loihi.Simulator(model, dt=dt, precompute=True) as sim:
    sim.run(length*dt)
```

But if I fix `tau=0.005` then there is no error.

hunse:
This makes them like Ensembles in this respect, and should avoid
problems with different taus on multiple inputs to an ensemble.

Fixes #94.

The other upside to this is that we should avoid the warning that we added for not having a filter on e.g. an input node.

The downside is that it gives the user less control over the filtering on nodes going into ensembles (there's always that 0.005 filter, though the length of that filter is configurable now if you know where to look). However, this same problem has always happened with ensembles to ensembles.


arvoelke:
Thanks for the quick fix! I'll try this branch out when I've got the chance.

> there's always that 0.005 filter, though the length of that filter is configurable now if you know where to look. However, this same problem has always happened with ensembles to ensembles.

Could you elaborate? Is this the additional filtering from interneurons? Where is this configured? If using full weights is this still a problem?

arvoelke:
Confirmed that this branch fixes the example in #94. However, if I then try to get rid of the interneurons on the recurrent connection by changing the solver to `weights=True`, the same exception comes up again:

```python
tau = 0.1

    u = nengo.Node(0)
    x = nengo.Ensemble(100, 1)
    nengo.Connection(u, x, synapse=tau)
    nengo.Connection(x, x, synapse=tau,

with loihi.Simulator(model) as sim:
    pass
```

```
~/CTN/nengo-loihi/nengo_loihi/simulator.py in __init__(self, network, dt, seed, model, precompute, target)
    182 
    183             # Build the network into the model
--> 184             self.model.build(network)
    185 
    186         self._probe_outputs = self.model.params

~/CTN/nengo-loihi/nengo_loihi/builder.py in build(self, obj, *args, **kwargs)
    129 
    130     def build(self, obj, *args, **kwargs):
--> 131         built = self.builder.build(self, obj, *args, **kwargs)
    132         if self.build_callback is not None:
    133             self.build_callback(obj)

~/CTN/nengo-loihi/nengo_loihi/builder.py in build(cls, model, obj, *args, **kwargs)
    156                 "Cannot build object of type %r" % type(obj).__name__)
    157 
--> 158         return cls.builders[obj_cls](model, obj, *args, **kwargs)
    159 
    160     @classmethod

~/CTN/nengo-loihi/nengo_loihi/builder.py in build_network(model, network)
    206     logger.debug("Network step 3: Building connections")
    207     for conn in network.connections:
--> 208         model.build(conn)
    209 
    210     logger.debug("Network step 4: Building probes")

~/CTN/nengo-loihi/nengo_loihi/builder.py in build(self, obj, *args, **kwargs)
    129 
    130     def build(self, obj, *args, **kwargs):
--> 131         built = self.builder.build(self, obj, *args, **kwargs)
    132         if self.build_callback is not None:
    133             self.build_callback(obj)

~/CTN/nengo-loihi/nengo_loihi/builder.py in build(cls, model, obj, *args, **kwargs)
    156                 "Cannot build object of type %r" % type(obj).__name__)
    157 
--> 158         return cls.builders[obj_cls](model, obj, *args, **kwargs)
    159 
    160     @classmethod

~/CTN/nengo-loihi/nengo_loihi/builder.py in build_connection(model, conn)
    706         mid_cx.add_axons(ax)
    707 
--> 708         post_cx.configure_filter(post_tau, dt=model.dt)
    709 
    710         if conn.learning_rule_type is not None:

~/CTN/nengo-loihi/nengo_loihi/loihi_cx.py in configure_filter(self, tau_s, dt, default)
    182         if self.decayU_set and not np.allclose(decayU, self.decayU):
    183             raise BuildError(
--> 184                 "Cannot change tau_s on already configured neurons")
    185 
    186         self.decayU[:] = decayU

BuildError: Cannot change tau_s on already configured neurons
```

hunse:
Yeah, so you're running into one of the constraints of the board. One of the reasons we use interneurons is that they allow individual filtering for each connection.

You can play around with interneurons by setting the parameters in `builder.py:Model`:
```python
loihi_model = nengo_loihi.builder.Model(dt=0.001)
loihi_model.inter_tau = 0.0  # remove interneuron filter
sim = nengo_loihi.Simulator(network, model=loihi_model)
```
(I haven't tested that, so there could be a typo)

However, that still won't let you use recurrent weight solvers, unless you set `inter_tau` to be the filter on your recurrent connection.

You could also do something like this:
```python
    u = nengo.Node(0)
    a = nengo.Ensemble(100, 1)
    x = nengo.Ensemble(100, 1)
    nengo.Connection(u, a, synapse=None)
    nengo.Connection(a, x, synapse=tau, 
    nengo.Connection(x, x, synapse=tau,
```

Ultimately, we could set up the `inter_tau` filter in your weights example so that it's just a default and can be overridden. In that case, though, the input would end up getting filtered twice (on the connection from the node to the interneurons, and from the interneurons to the ensemble), which might be surprising to users.

tcstewar:
> Yeah, so you're running into one of the constraints of the board.

I don't believe it's a constraint of the board -- it's a constraint of our current implementation.  This could be implemented on the board (I believe) by having a two-compartment neuron, each compartment getting a different tau, and just having pure addition between the two compartments.

arvoelke:
Could the example in my previous post somehow be mapped like:

```
u -> H(0) -> inter -> H(tau) -> x -> H(0) ----------.
                                ^                   |
                                '- H(tau) <- inter -'
```

excuse my weird drawing. But, basically put the taus on the outgoing interneuron filters, and set their incoming filters to zero? Happy to discuss why I'd like to do this offline, and just consider this kind of customization a feature request from a user. :)

tbekolay:
Also, in case anyone's trying to debug the learning differences introduced in this branch, for some reason performance is worse for larger numbers of neurons.

tcstewar:
> Once we get this figured out we can do a quick 0.3.1 or whatever.

Is there any possibility that we could rush out a quick 0.3.1 with this fix today?  Without it, I can't do many demos that use non-default synapses....    I can look into the `test_pes_comm_channel` and `test_node_no_synapse_warning` test failures today,

tbekolay:
I rebased this. Will do a temporary release (`v0.4.0.dev0`) which I will delete once we merge this and do a real release.

hunse:
> Could the example in my previous post somehow be mapped like:

```
u -> H(0) -> inter -> H(tau) -> x -> H(0) -----------.
                                ^                    |
                                '- H(tau) <- inter <-'
```

> excuse my weird drawing. But, basically put the taus on the outgoing interneuron filters, and set their incoming filters to zero? Moreover, could then use `weights=True` on the recurrent connection with the same filter. I think this is what you were saying above?

So something like that would be possible before this PR if you use a weight solver (though of course it wouldn't have the interneurons in the feedback connection).

The problem with doing things that way in general is that it then requires all connections to an ensemble to have the same `tau`. For example, you couldn't d a normal integrator network like this:
```python
with nengo.Network():
    u = nengo.Node([0])
    e = nengo.Ensemble(10, 1)
    nengo.Connection(u, e)
    nengo.Connection(e, e, synapse=0.1)
```
unless you set the synapse on the first connection to 0.1 as well (which we don't typically do).

xchoo:
@tbekolay asked me to post this here:
If you try to run the following network without #95, any jobs you run on the loihi hardware will hang at the  `lakemont_driver` call.

```
    stim = nengo.Node(1)
    ens = nengo.Ensemble(50, 1)

    # Comment out lines below to make it run, leave uncommented to make it hang
    stim_inh = nengo.Node(lambda t: t)
    nengo.Connection(stim, ens.neurons, transform=[[-2]] * ens.n_neurons)
```

hunse:
Ok, I've fixed up the learning stuff. The main issue was that the synapses in the test were different for different probes, such that the pre and post populations were being delayed much more than the stimulus. With that fixed, I could actually reduce the tolerances significantly.

I just removed the warning and the test in question, since with this new setup it's impossible to trip it.

hunse:
So I just tried Xuan's test. I found that it only hangs for me if I use a longer simulation time (if I run for 1s it hangs, but not for 0.1s).

Personally, I think it is something that would be good to look more into. But I don't think it's a good test. We don't really have any idea what it's testing, or if this PR actually fixes it or just shifts the problem (e.g. maybe even longer simulation times still make it hang, or more neurons, or something). I'd prefer to make it an issue and merge this as-is.

tbekolay:
> I just removed the warning and the test in question, since with this new setup it's impossible to trip it.

What does that mean? Setting `synapse=None` on that connection has no effect? I think that would be surprising to users, but I guess the node outputs are converted to spikes anyway which is a surprise. But yeah, what changed here such that this warning can no longer occur?

hunse:
Instead of doing node->no synapse->interneurons->synapse->ensemble, we now do node->synapse->interneurons->inter_tau->ensemble. Since only the second part of that pipeline is on the chip, and thus being build by the builder, with this PR all it ever sees is inter_tau synapses coming from Nodes. So they can never be none. The synapse is still done as the user would expect.

arvoelke:
Note the above code examples involving the setting of `inter_tau` no longer do the correct thing as of #132. Related discussion is in #97.

PR #98, opened 2018-09-28
 Clip weights during learning to emulate chip

hunse:
This helps make the emulator more like the chip, based on differences discovered in #89.

One question is where does the number 128 come from? I thought the weights were 8 bits plus a sign bit, which would give us a range of 256. However, 128 is the number that seems to replicate the behaviour.

Another question is do we want to clip or overflow? The chip appeared to have much weirder (noiser? rougher?) behaviour when the weights were nearing the threshold, which could indicate that they're overflowing instead of being nicely clipped. This would also fit the normal chip behaviour of overflowing instead of clipping.

It would be good to have a test that roughly compares the emulator and chip behaviour. This could be the test from #89 with the `function=lambda x: 0` added back in on the learned connection, since this is what illuminated the problem originally.


hunse:
Yeah, I think a warning would be good.

PR #103, opened 2018-09-29
Fix keyword spotting example

tbekolay:
@pblouw reported that his keyword spotting demo is broken in v0.3.0, and indeed [our keyword spotting example is broken also](https://www.nengo.ai/nengo-loihi/examples/keyword_spotting.html). He tracked it down to [this commit](https://github.com/nengo/nengo-loihi/commit/0811e2a840cf0c9bb24d77a10f43e6e3fc8a8486), which seems like it shouldn't even affect the keyword spotting network, but in fact it does.

The reason why that commit affects the keyword spotting network is that it translates **all** neuron->* connections in which `pre` is onchip and `post` is offchip with `transform=weights` to a `solver=NoSolver(weight.T)` connection. This would be fine if those two connections made no difference to how the network behaves, but that is unfortunately not the case, as evidenced by the keyword spotting example failing.

Investigating this issue is made more complicated by the fact that it is impossible to compare the two connection types because `transform=weights` connections are translated to `NoSolver` connections by the splitter. Commenting out this behavior in the splitter results in obvious differences between the two.

One way we can still investigate this is in looking at neuron->* connections in which `pre` is offchip and `post` is onchip, as this case was not handled in the same way as the reverse of that, so `transform=weights` connections still act differently from `solver=NoSolver` connections. This is definitely an oversight, but in this case a convenient one as it allowed me to write the `test_n2n_transform_solver` test in this PR.

Running that test with the commit it's introduced in (9042d86b6eeff64ac9079243087ddb0bd2233004) reveals some hard to debug issues. Thanks to the new `allclose` changes, here's the first bunch of failures when comparing a connections with a transform and a NoSolver:

```
allclose first 5 failures:
  (22, 0): 0.005813399449355831 -0.027141482844726717
  (23, 0): -0.027332068748132363 -0.026265229725634918
  (24, 0): -0.026442792216304224 -0.025404153073492673
  (25, 0): 0.051607908894409936 -0.024571305871791547
  (26, 0): 0.04992955092214089 0.05341804061865481
  (27, 0): 0.04829266554173085 0.051680339469823776
  (28, 0): 0.04670944364715465 0.0170311741195067
```

I attempted to fix this in splitter in f2ae792627a9f8250b98bbd2bd83dda4f5e85eac. I was not able to fix it (I find the `ChipReceiveNeurons` and `HostSendNode` stuff quite confusing), but looking at how it fails now gives me some hope:

```
allclose first 5 failures:
  (11, 0): -0.037291051016988146 0.0
  (12, 0): -0.03607528021115636 0.0
  (13, 0): 0.0058673941456823045 0.0
  (14, 0): 0.005681813349207329 0.0
  (15, 0): 0.0054955413512869184 0.0
  (16, 0): 0.005315376075829346 0.0
  (17, 0): -0.014805258804706747 -0.037291051016988146
```

It looks like the `NoSolver` case might be operating the same as the `transform` case, only 6 timesteps delayed. We should make them the same, but seeing as I tried to do that here, I think I will need help to do that.

Also, even if we fix this case, it doesn't fix the keyword spotting example. However, it might give us a bit of insight into why the switch from the `transform` connection to the `NoSolver` connection made a difference. Does the way we did the connection splitting in 0811e2a840cf0c9bb24d77a10f43e6e3fc8a8486 have the same weird issue as in 9042d86b6eeff64ac9079243087ddb0bd2233004, or is it simply delayed by some timesteps? My guess is that having several timesteps delay would not cause the keyword spotting example to change in the way that it changed, so I think we need to revisit the change in 0811e2a840cf0c9bb24d77a10f43e6e3fc8a8486 and ensure that switching the type of connection doesn't change behavior.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -22,7 +22,10 @@ Release history
 0.4.0 (unreleased)
 ==================
 
+**Changed**
 
+- An error is now raised if
+  a learning rule is applied to a non-decoded connection.
 
 0.3.0 (September 28, 2018)
 ==========================
--- a/nengo_loihi/builder.py
+++ b/nengo_loihi/builder.py
@@ -759,8 +759,7 @@ def conn_probe(model, probe):
         else:
             raise NotImplementedError()
 
-        target = nengo.Node(None, size_in=output_dim,
-                            add_to_container=False)
+        target = nengo.Node(size_in=output_dim, add_to_container=False)
 
         conn = Connection(probe.target, target, synapse=synapse,
                           solver=probe.solver, add_to_container=False,
--- a/nengo_loihi/splitter.py
+++ b/nengo_loihi/splitter.py
@@ -320,40 +320,27 @@ def split_chip_to_host(networks, conn):
         d=conn.size_mid,
         rng=np.random.RandomState(seed=seed))
 
-    if (isinstance(conn.pre, nengo.ensemble.Neurons)
-            and transform.ndim == 2):
-        # decoders manually specified in the transform
-        # should be handled like a normal decoder
-        probe = nengo.Probe(
-            conn.pre.ensemble,
-            synapse=None,
-            solver=nengo.solvers.NoSolver(transform.T),
-            add_to_container=False,
-        )
-        dims = transform.shape[0]
-        networks.chip2host_params[probe] = dict(
-            learning_rule_type=conn.learning_rule_type,
-            function=lambda x, dims=dims: np.zeros(dims),
-            transform=np.array(1),
-        )
-    else:
-        probe = nengo.Probe(
-            conn.pre,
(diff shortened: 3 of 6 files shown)


drasmuss:
See https://gitter.im/nengo/loihi-103 for some discussion points

drasmuss:
Rebased to #112 and reverted #85.  I pushed the old branch to https://github.com/nengo/nengo-loihi/tree/fix-ks-saved.  @tbekolay wrote a test and did some work there related to chip/host communication, but not directly related to the keyword demo, which we might want in the future.

tbekolay:
This LGTM but since I made the original PR I can't submit a review. Will find a scapegoat then merge.

PR #111, opened 2018-10-02
U/V overflow

hunse:
This has input current (U) and voltage (V) variables overflow in the emulator just like they do on the chip.

### DIFF
--- a/nengo_loihi/loihi_api.py
+++ b/nengo_loihi/loihi_api.py
@@ -25,6 +25,51 @@ V_MAX = 2**23 - 1
 V_MIN = -2**23
 
 
+def overflow_signed(x, bits=7, out=None, return_hits=False):
+    """Compute overflow on an array of signed integers.
+
+    For example, the chip uses 23 bits plus sign to represent U and V. We can
+    store them as 32-bit integers, and use this function to compute how they
+    would overflow if we only had 23 bits plus sign.
+
+    Parameters
+    ----------
+    x : array
+        Integer values for which to compute values after overflow.
+    bits : int
+        Number of bits, not including sign, to compute overflow for.
+    out : array (default: None)
+        Output array to put computed overflow values in.
+    return_hits : bool (default: False)
+        Whether to return a boolean array indicating which values overflowed.
+
+    Returns
+    -------
+    y : array
+        Values of x overflowed as would happen with limited bit representation.
+    o : array (if `return_hits`)
+        Boolean array indicating which values of `x` actually overflowed.
+    """
+    if out is None:
+        out = np.array(x)
+
+    smask = 2**bits
+    xmask = smask - 1
+
+    omask = ~(smask - 1)
+    overflowed = (out & omask) != 0
+
+    sign = out & smask
+    signbase = ~(sign - 1)  # if sign bit, turn all bits left of sign bit on
+    out &= xmask
+    out += signbase
+
+    if return_hits:
+        return out, overflowed
+    else:
+        return out
+
+
 def vth_to_manexp(vth):
     exp = VTH_EXP * np.ones(vth.shape, dtype=np.int32)
     man = np.round(vth / 2**exp).astype(np.int32)
--- a/nengo_loihi/loihi_cx.py
+++ b/nengo_loihi/loihi_cx.py
@@ -11,6 +11,7 @@ from nengo.utils.compat import is_iterable, range
 from nengo_loihi.loihi_api import (
     BIAS_MAX,
     bias_to_manexp,
+    overflow_signed,
     SynapseFmt,
     tracing_mag_int_frac,
     U_MAX, U_MIN,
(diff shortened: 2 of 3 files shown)


nengo_loihi/loihi_cx.py
```diff
@@ -686,24 +686,26 @@ def step(self):  # noqa: C901
...
         self.v[:] = self.decayV_fn(self.v, u2)
-        if np.any(self.v > V_MAX):
-            self.error("Overflow in V (max was %d)" % self.v.max())
-        if np.any(self.v < V_MIN):
-            self.error("Underflow in V (min was %d)" % self.v.min())
-        self.v = np.clip(self.v, a_min=V_MIN, a_max=V_MAX, out=self.v)
+        # _, o = overflow_signed(self.v, bits=V_BIT, out=self.v,
+        #                        return_hits=True)
+        # if np.any(o):
+        #     self.error("Overflow in V")
```
drasmuss: Why is this part commented out?  Do we not need to check for overflow in V?  Or is this related to the comment in the test, that you couldn't find a way to get V to overflow?

nengo_loihi/loihi_api.py
```diff
@@ -19,10 +19,56 @@
...
-U_MAX = 2**23 - 1
-U_MIN = -2**23
-V_MAX = 2**23 - 1
-V_MIN = -2**23
+Q_BIT = 21  # number of bits for synapse accumulator
+U_BIT = 23  # number of bits for cx input (u)
+# V_BIT = 24  # number of bits for cx voltage (v)
+
+
+def overflow_signed(x, bits=7, out=None, return_hits=False):
```
drasmuss: I think we need some unit tests for this function.  It seems like there are a lot of different conditions to worry about, and a lot of bit math with potential to go awry.  `test_uv_overflow` checks that the emulator matches the hardware given certain inputs, but it's hard to know whether that is covering all the potential overflow cases.  In addition, it would be nice to assert concrete behaviour (e.g. given these inputs, this is what the output should be), rather than just matching to the hardware.  That way we know that the overflow behaviour is working exactly the way we think it is, as opposed to having to wonder whether some mismatch is due to a problem in our emulator, a problem in the hardware, or a problem in this function.

nengo_loihi/loihi_cx.py
hunse: Exactly. I feel like it should overflow, and that this code should thus be in there, but I couldn't actually figure out a way to hit this and test the code, so I commented it out.

nengo_loihi/loihi_api.py
hunse: Any ideas on how to write a unit test for this? I'm not quite sure how to write a general test for it, without just replicating the code in the function. I guess I could just write it without all the bit twiddling optimizations that I put in.

drasmuss: You just want to test specific cases that cover the different overflow possibilities.  Something like 
```
# basic check
assert(overflow_signed(16, bits=2) == ?
# check that negative values are handled appropriately
assert overflow_signed(-16, bits=2 == ? 
# check that changing the value of bits changes the overflow as we expect
assert overflow_signed(16, bits=3) == ?
...
```
and so on to check other possibilities (e.g. other values for x? is zero handled correctly? is `return_hits` correct in all these cases?).  You probably know better than me the different ways in which these values could overflow, and what edge cases we'd need to worry about.

hunse: Ok, I can do something like that. In my experience, though, it's hard to predict the edge cases without having a robust programmatic test that tests it for thousands/millions of potential values (you never know when you'll hit something like the 300 bug). But I'll put in what I can think of.

hunse:
Ok, I've written a test. In writing it, I also figured out how to simplify the overflow function significantly (and make it faster), so that's good!

nengo_loihi/tests/test_loihi_api.py
```diff
@@ -1,23 +1,24 @@
 import numpy as np
+import pytest
 
 from nengo_loihi.loihi_api import overflow_signed
 
 
-def test_overflow_signed():
-    for b in (8, 16, 17, 23):
-        x = np.arange(-2**(b-2), 2**(b+2), dtype=np.int32)
+@pytest.mark.parametrize("b", (8, 16, 17, 23))
```
drasmuss: Made this a parameter, as it makes things slightly easier to debug if one of these values starts failing.

nengo_loihi/tests/test_loihi_api.py
```diff
@@ -1,23 +1,24 @@
...
+def test_overflow_signed(b):
+    x = np.arange(-2**(b+2), 2**(b+2), dtype=np.int32)
```
drasmuss: I think this needs to be a `+` in order to test negative overflow.

drasmuss:
Actually I'm just looking at the `codecov` diff and none of the overflow checks in `CxSimulator` are being hit, which makes me think that `test_uv_overflow` is not testing what it seems to be testing?

nengo_loihi/tests/test_loihi_api.py
hunse: Yes! Thanks for catching that.

nengo_loihi/loihi_cx.py
tbekolay: I am pretty sure that I've gotten an overflow in V in the Nengo tests... I'll try to track it down and see if it can be distilled into a test.

hunse:
I just realized that this doesn't work with the simulator `simreal` target. I'm not sure if we have any tests for this right now. I can look at adding a few tests and getting this working with `simreal`, unless anybody else has time on their hands and is looking for a project.

tbekolay:
Hm, I think I wouldn't block this PR on `simreal` but it would be nice to do later on. Perhaps make a separate issue for it? It would be good to add some other tests for simreal too.

Issue #20, opened 2018-08-20
Better rules for splitter


tbekolay:
In our PES learning example, I had to mark two ensembles as being simulated off-chip because the splitter wanted to put them on the chip. It's useful, I think, to always have the ability to mark ensembles as on- or off-chip, but it shouldn't be necessary for a simple PES learning example. I think modifying the splitter to work on connections first (instead of last) will make it possible to detect the learning rule case. Or, we might have to do multiple passes over the network.

drasmuss:
Resolved in #112

PR #112, opened 2018-10-02
Refactor splitter

tbekolay:
Attempting to make each rule done in the splitter a separate unit, making it easier to read and test. The units are ordered so that the overall behavior is well defined.

Still a work in progress, but making the PR now as the overall structure is there, just finishing converting over the rest of the `split` function.

### DIFF
--- a/docs/examples/learn_communication_channel.ipynb
+++ b/docs/examples/learn_communication_channel.ipynb
@@ -48,8 +48,6 @@
    "outputs": [],
    "source": [
     "with nengo.Network(label=\"PES learning\") as model:\n",
-    "    nengo_loihi.add_params(model)\n",
-    "\n",
     "    # Randomly varying input signal\n",
     "    stim = nengo.Node(WhiteSignal(60, high=5), size_out=1)\n",
     "\n",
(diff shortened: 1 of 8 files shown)


tbekolay:
This PR's now in what I would consider a "first draft" stage -- the code organization is essentially what it should be, but it currently doesn't work (splits everything to the host). I'm going to start writing tests next, as that was a big part of why it's organized this way, and should make it clear why everything's being split to the host.

tbekolay:
This PR now works in that all of the unit tests that were previously passing still pass, and with approximately the same speed as before (possibly faster). It doesn't change as much as I wanted to change (e.g., removing `chip2host_params`, `chip2host_receivers` and `host2chip_senders`; moving more of the splitter to the builder; removing the node subclasses in favor of processes), but the code is easier to read and test, and it is now easy to investigate what the splitter did by looking at the `adds`, `moves`, and `removes` attributes of `SplitNetworks`.

I am still working on a few things:

- [ ] Finishing up the tests
- [ ] Ensuring that this PR fixes #20 by removing the manually specified ensemble placement

While I'm finishing these things up, people who are interested in reviewing can look at [the refactored `splitter.py`](https://github.com/nengo/nengo-loihi/blob/fb8ccb6cb7cc79c77693e4450cb4a4683dbca198/nengo_loihi/splitter.py).

tbekolay:
This PR is now ready to be reviewed; I finished writing tests, ensured that all old tests passed, and modified any learning networks that were previously manually setting `.on_chip = False` in order to run to not do that, and those tests still pass. The PES example in the docs is also modified and works as it did before.

Requesting that @tcstewar review since he wrote the original splitter code to ensure I didn't miss anything.

PR #113, opened 2018-10-03
Emulator input delay

hunse:
On Loihi, spikes always take one timestep to get where they're going. With input spikes in the emulator, we were not accounting for that delay and having them arrive instantaneously. This PR fixes that.

Based off #111, so that I could update the test there to get rid of the delay.


hunse:
This is a small change, so I'm hoping @tbekolay can do both the review and merge.

PR #117, opened 2018-10-05
Cifar basic

hunse:
A basic 1-channel network for CIFAR-10.

Currently trains OK for using only one channel (~40% error), but then does not perform well in the emulator. Not sure if it's worth pursuing this further, or focusing on the multi-channel version instead.


celiasmith:
if it will be easier to debug one channel in the emulator and that carries over to 3, i'd do the 1... (ie if 1 chan *has* to work for 3 to work, might as well debug the simpler network)

tbekolay:
Superseded by the now merged #161 and https://github.com/nengo/nengo-examples/pull/19.

PR #119, opened 2018-10-11
Modifying tau_s gives a warning, not an Exception

tcstewar:
Just a quick change, but as discussed in the meeting it might be worth addressing the fact that this makes the final model dependent on the build order (as it uses the final tau_s value).  Eventually we will support different tau_s values, but for now perhaps we should take the one with the longest time constant?

### DIFF
--- a/nengo_loihi/loihi_cx.py
+++ b/nengo_loihi/loihi_cx.py
@@ -42,8 +42,8 @@ class CxGroup(object):
         Input (synapse) decay constant for each compartment.
     decayV : (n,) ndarray
         Voltage decay constant for each compartment.
-    decayU_set : bool
-        Whether decayU has been explicitly set
+    tau_s : float or None
+        Time constant used to set decayU. None if decayU has not been set.
     scaleU : bool
         Scale input (U) by decayU so that the integral of U is
         the same before and after filtering.
@@ -92,7 +92,7 @@ class CxGroup(object):
 
         self.decayU = np.ones(n, dtype=np.float32)  # default to no filter
         self.decayV = np.zeros(n, dtype=np.float32)  # default to integration
-        self.decayU_set = False  # whether decayU has been explicitly set
+        self.tau_s = None
         self.scaleU = True
         self.scaleV = False
 
@@ -162,7 +162,20 @@ class CxGroup(object):
         assert probe.target is self
         self.probes.append(probe)
 
-    def configure_filter(self, tau_s, dt=0.001, default=False):
+    def configure_default_filter(self, tau_s, dt=0.001):
+        """Set the default Lowpass synaptic input filter for Cx.
+
+        Parameters
+        ----------
+        tau_s : float
+            `nengo.Lowpass` synapse time constant for filtering.
+        dt : float
+            Simulator time step.
+        """
+        if self.tau_s is None:  # don't overwrite a non-default filter
+            self._configure_filter(tau_s, dt=dt)
+
+    def configure_filter(self, tau_s, dt=0.001):
         """Set Lowpass synaptic input filter for Cx to time constant tau_s.
 
         Parameters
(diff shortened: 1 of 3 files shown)


tbekolay:
I would be OK with taking the longest time constant; if this is indeed overwriting the `inter_tau`, then another option would be to replace it iff the new value is not `inter_tau`, as presumably the user wants the one they specified.

We also need to change the warning message, but what we change it to depends on the intent here. So if we are overwriting `inter_tau`, then we should warn about that. If we're taking the longest, then we should say that.

tbekolay:
I added a commit only overwriting `tau_s` if the new `tau_s` is longer than the old one. There are now two warnings, one if it's overwritten and one if it's not, and a test that ensures that these warnings are displayed in those situations.

However, one issue for this commit is that five tests have started failing. These tests are all ones which explicitly set `synapse=None` on some connection, resulting in a `tau_s` of 0 (and `decayU` of 1). Since the default `tau_s` is 0.005, it is now impossible to set `tau_s` to a lower value.

One way to get around this would be to always allow overriding if the original value was the default. However, currently the way in which we set the default is through a kwarg to `configure_filter`, so outside of this method it's impossible to tell whether the current `tau_s` value was explicitly set or set as the default. Adding a new attribute that tracks whether the value set is the default value or a new value really feels bad to me... we're trying to keep the same interface for setting the value and for setting the default value, but if we need to also add in an attribute to track whether it was set as default or not, then it feels like we shouldn't be trying to keep the interface the same.

@hunse since you recently changed `configure_filter`, what do you think the correct resolution is? I'm thinking that there should be separate `configure_default_filter` and `configure_filter` methods, and we should explicitly track the default.

nengo_loihi/loihi_cx.py
```diff
@@ -173,19 +173,24 @@ def configure_filter(self, tau_s, dt=0.001, default=False):
...
+            return
+        elif self.tau_s is not None and tau_s > self.tau_s:
+            warnings.warn(
+                "tau_s is currently %g, which is smaller than %g. Overwriting "
+                "tau_s with %g." % (self.tau_s, tau_s, tau_s))
 
+        decayU = 1 if tau_s == 0 else -np.expm1(-dt/np.asarray(tau_s))
         self.decayU[:] = decayU
         self.scaleU = decayU > self.DECAY_SCALE_TH
-        self.decayU_set = not default
```
hunse: So the idea behind the `default` value is that it initializes `decayU`, but in a way where we won't get mad if that value is overridden later. I'm worried that now, we're going to throw more warnings, because we're going to warn even if the user overwrites the default. I don't think we should do this, but I agree, it is annoying to track whether it has been set. Perhaps if `default=True`, we don't set `tau_s`, we only set `decayU`. Then, if something overwrites that default, we won't complain. Does that make sense?

hunse: Just reading your comment above. This would solve those errors, too.

tbekolay: That third paragraph in the comment above is essentially a response to that question... I was going to only set `self.tau_s` if we're not setting the default, but then we can't tell if the default has been set yet (because right now we check `if self.tau_s is not None`). Is there another way to tell if the default's been set?

hunse: Why do we need to know if the default has been set? We didn't need to know this before. The point is that it's just a default, and we don't care if it gets overridden.

tbekolay: Previously we weren't allowing non-defaults to be overridden, it would just raise an exception, so there was only one case: overriding a default. Now, there are two cases, overriding a default, and overriding a value explicitly set, and we need some way of differentiating between those two.

tbekolay:
This PR just keeps on giving... so with those recent changes, `test_tau_s_warning` has changed and now fails. What happens now might be the correct behavior, but I want to put it here to make sure I understand what's going on.

This network:

```python
with nengo.Network() as net:
    stim = nengo.Node(0)
    ens = nengo.Ensemble(10, 1)
    nengo.Connection(stim, ens)
    nengo.Connection(ens, ens,
                     synapse=0.001,
```

Gives

```
UserWarning: tau_s is currently 0.001, which is smaller than 0.005. Overwriting tau_s with 0.005.
```

This seems strange to me, because I think that the 0.005 is coming from the `stim` to `ens` connection, which comes after the recurrent connection.

Adding another connection makes this even stranger:

```python
with nengo.Network() as net:
    stim = nengo.Node(0)
    ens = nengo.Ensemble(10, 1)
    nengo.Connection(stim, ens)
    nengo.Connection(ens, ens,
                     synapse=0.001,
    nengo.Connection(ens, ens,
                     synapse=0.1,
```

This gives

```
UserWarning: tau_s is currently 0.001, which is smaller than 0.1. Overwriting tau_s with 0.1.
```

Which makes sense, but then

```
UserWarning: tau_s is already set to 0.1, which is larger than 0.005. Using 0.1.
```

If we remove the `stim` connection...

```python
with nengo.Network() as net:
    ens = nengo.Ensemble(10, 1)
    nengo.Connection(ens, ens,
                     synapse=0.001,
    nengo.Connection(ens, ens,
                     synapse=0.1,
```

Then it makes sense

```
UserWarning: tau_s is currently 0.001, which is smaller than 0.1. Overwriting tau_s with 0.1.
```

I am thinking that the `stim` connection happens last because of the splitter changing the order of things ... are we OK with this?

hunse:
> I am thinking that the stim connection happens last because of the splitter changing the order of things

Yes, that appears to be the case.

I am fine with this. It looks like we made a good decision to go with taking the largest value, since the behaviour is the same which ever way they get added.

tbekolay:
> It looks like we made a good decision to go with taking the largest value, since the behaviour is the same which ever way they get added.

Good point! I'll modify the test accordingly then merge.

PR #121, opened 2018-10-11
Close sub-simulators on close

### LINKED ISSUE #102: Simulator.close should also close split networks
Right now we get `ResourceWarning`s related to not calling `self.host_sim.close()` etc.

drasmuss:
Also adds a close method to the emulator (for consistency).

Fixes #102

TODO:
- [ ] figure out why stopping the n2driverservice is giving an error when running on Loihi


nengo_loihi/loihi_cx.py
```diff
@@ -496,6 +496,7 @@ def __init__(self, model, seed=None):
 
         self._probe_filters = {}
         self._probe_filter_pos = {}
+        self.closed = False
```
tbekolay: Safest to put this line first, before `self.build`, in case we ever make this a context manager in the future (we had some instances in `nengo.Simulator` where we would be looking at `self.closed` in the `__exit__`, which gave an attribute error).

nengo_loihi/simulator.py
```diff
@@ -273,8 +273,6 @@ def __enter__(self):
         return self
 
     def __exit__(self, exc_type, exc_value, traceback):
-        if self.loihi is not None:
-            self.loihi.__exit__(exc_type, exc_value, traceback)
```
hunse: We call `self.loihi.__enter__` in the `__enter__` function for this class, so I think we need to call `__exit__` too rather than just `close` (or else call neither).

If we call neither, then we're basically making the assumption that `LoihiSimulator.__enter__` does nothing, and `LoihiSimulator.__exit__` just calls `close`. I'm okay with this assumption, since I think it's generally assumed with nengo simulators of all types (for example, it's typically assumed that making a simulator and calling `close` when you're done is the same as using the context manager, though of course we encourage the context manager for its exception handling).

nengo_loihi/loihi_cx.py
```diff
@@ -780,3 +782,6 @@ def get_probe_output(self, probe):
         x = np.asarray(self.probe_outputs[cx_probe], dtype=np.float32)
         x = x if cx_probe.weights is None else np.dot(x, cx_probe.weights)
         return self._filter_probe(cx_probe, x)
+
+    def close(self):
+        self.closed = True
```
hunse: I'm not sure if I agree with adding this `close` that does nothing. My problem with it is we haven't done all the other legwork to make sure `CxSimulator` gets closed (when we create it manually in some of our tests). If someone actually adds something to this function down the line, they might not realize all that legwork still needs to be done.

If we want this, I'd add a context manager, make sure it's used in tests, and make sure we error when trying to run a closed simulator. I also might try to free up some of the memory allocated in `build`. I'm happy to do this if others agree.

drasmuss: Doesn't seem super high priority to me (we don't expect users to be manually creating `CxSimulator`, and we only do that in one test, so I think allowing `nengo_loihi.Simulator` to handle all the opening and closing is fine).  But doesn't hurt I guess, if you're looking for something to work on 🤷‍♂️

drasmuss:
Fixups look good to me.

drasmuss:
Pushed a commit that stores all the sub-simulators in a dictionary, rather than as attributes of the parent simulator.  Makes it easier to manage common tasks for those simulators, and I think it makes it clearer what those variables are representing to in the code (e.g. `self.sims["loihi"]` vs `self.loihi`).

However, in making those changes I noticed that we seem to have some dead code branches (I noticed because there were checks for `if self.host_sim is None`, but `self.host_sim` could never be `None`).  See https://github.com/nengo/nengo-loihi/pull/121/commits/587dcee8a0ebf6714bb8cf450924cc7758fc553d (which passes all the unit tests on emulator and loihi).  I'm not positive what the intended use case was for those branches, so I'm not sure what the best resolution is.  Should we add tests to cover those branches? Or just remove those branches?

tbekolay:
> Should we add tests to cover those branches? Or just remove those branches?

I think in the refactoring commit I refactor the run code anyway, so maybe I'll just apply those changes here and add some tests, will make the refactoring PR a bit easier down the line.

nengo_loihi/simulator.py
```diff
@@ -417,6 +417,7 @@ def run_steps(self, steps):
                     self.sims["emulator"].step()
                     self.handle_chip2host_communications()
             else:
+                raise Exception
```
hunse: What's this exception for? Same down below.

hunse:
Changes look good to me.

tbekolay:
Pushed two fixups finishing the last few tests and fixing issues from those tests. With those, I'm satisfied with this branch! @hunse could you please re-review.

nengo_loihi/simulator.py
hunse: Ahh, because these never get hit. In the future, I would put that in the exception comment. These get removed by a later commit, though, so not worth fixing here.

nengo_loihi/simulator.py
```diff
@@ -487,7 +488,25 @@ def run_steps(self, steps):
...
+                for _ in range(steps):
+                    h2c.write(h2c.numElements, [0] * h2c.numElements)
+                    c2h.read(c2h.numElements)
+                    if not printed:
+                        print(traceback.format_exc())
+                        print("\nAttempting to end simulation...")
+                        printed = True
+                self.sims["loihi"].n2board.nxDriver.stopExecution()
+                self.sims["loihi"].n2board.nxDriver.stopDriver()
+                self.sims["loihi"].wait_for_completion()
```
hunse: Oh, I forgot to say: even though it doesn't make a difference, I think we should call this before we stop excecution and the driver. Intel said that's the way it should go, so I would air in that direction. Also, I think the `n2board` calls should be done by the Loihi simulator, maybe in `disconnect`.

nengo_loihi/loihi_interface.py
```diff
@@ -601,7 +600,7 @@ def _host2chip_spikes(self):
     def _host2chip_errors(self):
         errors = []
         for sender, receiver in self.model.host2chip_senders.items():
-            if isinstance(receiver, PESModulatoryTarget):
+            if hasattr(receiver, "target"):
```
hunse: Is there a reason behind this change? I feel like a lot of things could have a `target` attribute, but this code is very specific to PES learning (i.e. the errors are passed to a snip that targets them specifically for PES), so I don't see any reason to be general here.

nengo_loihi/simulator.py
```diff
@@ -184,9 +184,17 @@ def __init__(self, network, dt=0.001, seed=None, model=None,  # noqa: C901
...
             if len(self.host.all_objects) > 0:
                 self.sims["host"] = nengo.Simulator(
                     self.host, dt=self.dt, progress_bar=False)
+            elif not precompute:
+                # If there is no host and precompute=False, then all objects
+                # must be on the chip, which is precomputable in the sense that
+                # no communication has to happen with the host.
+                # We could warn about this, but we want to avoid people having
+                # to specify `precompute` unless they absolutely have to.
+                self.precompute = True
```
hunse: I don't like this. If I say `precompute=False`, I want that to be followed. Right now, there are all kinds of bugs in `precompute=True`. For example, it doesn't work well with large networks, in my experience as well as that of others. Though I think that might have to do with spike generation, so maybe it's a non-issue for networks without a host sim. Anyway, I think this change might be good long-term, when we've really sorted things out with `precompute`, but right now I'd rather things be dumb and just do what I ask.

Another option would be to have `precompute=None` as the default, in which case I would be fine with something like this, since then I can always say `precompute=False` explicitly and force it to be false.

nengo_loihi/simulator.py
hunse: I put up a fix for this, and found that I can actually get rid of the hang on exit as well, so that when we get an exception, the simulation exits as we would expect.

Unfortunately, there are still some hangs (for example, with the learning tests), that happen in a different place and aren't solved by this, but they weren't solved before my fixup, either.

hunse: I should mention, that I think the learning hang is because we don't combine errors properly when we have more than one input to a learning rule. Rather than summing the inputs, we concatenate them so we end up with too many error values that we then try to send to the chip, and I think this causes the hang. So that one might not a Python exception at all, so it's not surprising that nothing in this branch helps with it.

hunse:
All the Loihi tests pass for me now.

tbekolay:
@hunse's changes LGTM, will merge on his approved review (next week, when I'm back from Toronto)

hunse:
I still have two questions, about checking for `target` instead of a PES-specific check, and about setting `precompute=True` in the case that there's no host sim.

hunse:
I found one other issue with this. Now, because we loop through the model probes in LoihiSimulator.build, the `cx_model` that's passed in needs to be a `builder.Model` so that it has the `probes` attribute; it can no longer be a general `CxModel` (assuming `use_snips=True`). I use this in one of my tests in the `cifar` branch.

Ultimately, this is a place where a better mapping between Nengo objects, loihi_cx objects, and loihi_interface objects would be very helpful. Until we have that, an immediate quick fix isn't apparent to me, but I can look at this more on Monday and see if I can figure anything out.

hunse:
Ok, I think this commit should get things working with general CxModels again. Let me know what you think @tbekolay. If you're happy with it, I'm happy to squash it into yours on merge, I just kept it separate for now.

hunse:
Okay, I've done more refactoring off this branch so that CxSimulator and LoihiSimulator work with general CxModels (including using snips, getting probe data, etc.). This is all in the `close-sims-hunse` branch.

The only thing is I've based that work off #133. I'm thinking I'll pull what's directly pertinent to this PR here, and make a separate PR for the big refactoring.

hunse:
Okay, I've put the pertinent stuff here. I'll make a PR down the road with the refactoring that's currently in `close-sims-hunse`.

My changes end up being quite minimal. Basically, I added a note about a test failure, and changed the exception printing so that it only happens if we're using snips. Otherwise, we pass the exception as normal.

nengo_loihi/simulator.py
```diff
@@ -488,22 +488,23 @@ def run_steps(self, steps):
...
-                    if not printed:
-                        print(traceback.format_exc())
-                        print("\nAttempting to end simulation...")
-                        printed = True
+                self.sims["loihi"].wait_for_completion()
                 self.sims["loihi"].n2board.nxDriver.stopExecution()
                 self.sims["loihi"].n2board.nxDriver.stopDriver()
-                self.sims["loihi"].wait_for_completion()
+            else:
+                raise e
```
tbekolay: You can actually just do `raise` here without `e` and without the `as e` above and it'll do the same thing

hunse: Oh, cool, that's good to know!

nengo_loihi/loihi_interface.py
tbekolay: Yes, so because the check for `PESModulatoryTarget` got moved from `simulator.py` to `loihi_cx.py` and `loihi_interface.py` there is now a cyclic dependency between `splitter.py` and `loihi_cx.py` because of the `PESModulatoryTarget` check. The cyclic dependency means I needed to check without importing `PESModulatoryTarget` in `loihi_cx.py`, so I made the check the same in `loihi_interface.py` too. I agree that it's not the best check, but `PESModulatoryTarget` is also kind of a weird class that could be replaced by a dict or something else, so I think that will get refactored in the near future anyhow.

tbekolay: But yeah, if you prefer a different way to doing the check without importing `PESModulatoryTarget` that's probably fine too (e.g. checking `type(receiver).__name__` or whatever)

nengo_loihi/simulator.py
tbekolay: Like you say, I think the `precompute=True` bugs that we've encountered are irrelevant here because this case only occurs if there is nothing running on the host, meaning that we are not generating any spikes to inject into the model, we are simply letting Loihi run on its own with no interaction from us, then reading the values off the chip later. It seems silly to slow down the chip for no reason other than that we're nervous there might be bugs.

Speed argument aside, the main reasons I made this change is that it was a silly amount of work to get rid of the complexity warning for `run_steps`, requiring making a `_make_run_steps` method that then calls either `_make_emu_run_steps` and `_make_loihi_run_steps`, which both just barely pass the complexity check. If we allow `precompute=False` unequivocally, then we have another case in both of those methods, so they will fail the complexity check and we have to figure out another way to break up these methods. So while I agree that passing `precompute=None` and then determining it if not specified would be better, it mean that the `precompute=False` and no host case is possible, so we have to implement it regardless.

I don't disagree that explicitly setting `precompute=False` and having it change on you would be annoying, but I really can't imagine anyone being extremely opinionated about making sure their model is not precomputed while also having no nodes in their model. So, I guess I would be okay with setting `precompute=None` and then raising a `NotImplementedError` for having `precompute=False` and no `host`, because that would be a use case I think we would want to know about.

tbekolay:
Responded to your comments @hunse, I'll wait for your response before rebasing and cleaning up the history (as it's getting to be too many commits).

nengo_loihi/loihi_interface.py
hunse: Typically when I run into these circular dependencies I just put the import at the top of the function. It's not ideal, but as long as one of our unit tests checks it I think that should be fine (as I understand it, the reason for putting imports at the top of a file is so that if they fail, they fail fast. If the imports are intra-repository, I'm less worried about this.) How would you feel about that @tbekolay?

nengo_loihi/simulator.py
hunse: OK, I'm fine keeping this for now. I agree that it does seem to be kind of silly just to avoid it because there could be bugs. If we do encounter any bugs, then we can revisit this. I just wish we knew more about why we're having issues with `precompute=True` on large models. I think the working hypothesis that it's due to the number of input spikes is still the best, but I don't think anybody has looked at this in detail.

nengo_loihi/loihi_interface.py
tbekolay: I really dislike the imports within functions. IMO the imports at the top of the file are there so the reader can see what this file depends on, i.e., how it interacts with the rest of the codebase (and external libraries). So when another dependency gets introduced in the middle of the file it breaks my mental model of where this file sits in relation to the other files in the repository. The interdependencies between files in a source tree is a big part of my mental model of how the code is organized.

nengo_loihi/simulator.py
tbekolay: My hope is that we can speed up `precompute=False` and get rid of the precomputing logic before we have to look into it in detail :crossed_fingers:

nengo_loihi/loihi_interface.py
tbekolay: I would be happy with moving `PESModulatoryTarget`, `HostSendNode`, `HostReceiveNode`, etc (basically the parts of the splitter after `split_pre_from_host`)... perhaps to `loihi_cx.py`? Or `loihi_api.py`? They'll be moved in a later refactoring anyhow so it's not too critical, but might as well get agreement on it now.

hunse: I would maybe suggest something like `communication.py`, since all these have to do with host-chip communication. But maybe it's not worth adding another file if it's just going to get changed in refactoring. They don't really fit in either `loihi_cx` or `loihi_api`, though. Where do they go in the refactoring?

hunse: Or could they go in `builder.py`?

tbekolay: They stay in the splitter, but if we're moving them then they'll have to move again in the refactoring, is what I mean. What these nodes are doing should eventually be put into the builder anyhow, so maybe `builder.py`?

hunse: I hate having conversations on Github, because it really doesn't refresh the messages...

hunse: Realized that in `builder.py`, we create a circular dependency because it needs `loihi_cx.py`, which needs `PESModulatoryTarget` in `builder.py`. @tbekolay and I talked about this a bit more on Slack, and decided to put them in `loihi_cx.py` temporarily. They'll get moved again in the upcoming refactoring, so just deferring the decision until then.

tbekolay:
Rebased and squashed the history; this looks good to me, so I'm going to check each commit again and run on hardware then merge. Will probably take an hour or so, so @drasmuss and @hunse if you have any objections that's your window ;)

Issue #99, opened 2018-09-28
Teach users about initial PES functions


tbekolay:
As we found out in #89, the choice of the initial function across a PES learned connection can affect its behavior significantly. While we should determine more robust ways to discretize in the long term (see #83), in the short term we should at least attempt to inform users that this choice makes a big difference.

A few ideas on how we might do this:

- Make an example showing how different things can be given the initial function (similar to what's in the #89 comment thread)
- Raise a warning either at model construction or model build time if the user has made a learned connection and specified a function. This is similar to what we're doing with connection from nodes with `synapse=None` right now. This assumes that the identity function (communication channel) always puts weights in a decent range, which might not be true.

drasmuss:
I believe the discretization has been improved so that using an all-zero initialization no longer causes things to fall apart, @hunse can confirm?

hunse:
Yep. We now test initializing with zeros, too: https://github.com/nengo/nengo-loihi/blob/master/nengo_loihi/tests/test_learning.py#L37

PR #122, opened 2018-10-11
Nengo benchmarks

tcstewar:
I'm not sure if this should get merged into master, but I'm opening this PR anyway if people want to discuss the benchmark document.


tcstewar:
> Running the notebook I got as far as the LearningSpeedup example, then I get

That's very odd....  I don't get that at all...  I'll try on a few other machines to try to replicate it...

drasmuss:
It'd be good to note in the notebook which benchmarks are expected to match well, and which don't (and maybe an explanation for why they don't match if we know).  Just from visual inspection I would say the ones that don't match for me are

- Circular Convolution (rmse 0.1 vs 0.38)
- Spa sequence (slower cycle time)
- Semantic memory (memory -0.01 vs 0.33)
- Semantic memory with recall (memory 0.0 vs 0.5)
- Oscillator (graphs are ~10x different scales)
- Spa sequence routed (getting stuck in cycle)
- Cleanup (recall strength 0.2 vs 0.9)

But I'm not sure which of those are expected or not.

tcstewar:
> Running the notebook I got as far as the LearningSpeedup example

I've now tried this on three different machines and I can't replicate it...  it runs fine for me.  :(  I'm not sure how to try to debug that....



> It'd be good to note in the notebook which benchmarks are expected to match well, and which don't

I've added a description for all of them now, and categorized them into Passing, Non-passing, and Non-running.  All of the non-passing ones seem to be the integrator issue, and the non-running one works fine in the emulator, but uses too many axons on the chip.

tcstewar:
> * It would be good if this worked with the released version of pytry (or do a new pytry release)

Ooo, good idea.  It does work with the current release of pytry -- it's just that it ignores the resizing of the figure so it looks a little uglier.  I'll look into doing a release, and possibly also dusting off that PR that makes nengo_benchmarks not dependent on pytry....

tcstewar:
I should also take a stab at making the figures more informative -- that'd be a good thing to do anyway as as we use `nengo_benchmarks` for different things....

drasmuss:
> I've now tried this on three different machines and I can't replicate it... it runs fine for me. :( I'm not sure how to try to debug that....

It happens 100% of the time for me 🤷‍♂️ .  How are the benchmarks seeded, could that maybe be different between machines?

> I've added a description for all of them now, and categorized them into Passing, Non-passing, and Non-running. All of the non-passing ones seem to be the integrator issue, and the one non-running one works fine in the emulator, but uses too many axons on the chip.

That organization looks great, definitely helps frame what users should expect to see on each one.

> It does work with the current release of pytry

It didn't work for me, some error about missing `plt` argument

tcstewar:
> It happens 100% of the time for me 🤷‍♂️ . How are the benchmarks seeded, could that maybe be different between machines?

Hmm, possibly... they're seeded by setting both the python random seed and the numpy random seed:
https://github.com/tcstewar/pytry/blob/master/pytry/trial.py#L81

I'm not sure what situations would allow those to give different generated sequences....   Does the problem go away if you switch seeds (by specifying seed=X in the dictionary)?



> It didn't work for me, some error about missing `plt` argument

Grrr, I tested on the wrong version....   Sorry about that.  I'll do a release, and add both pytry and scikit-learn to the setup.py for nengo_benchmarks....

tcstewar:
> > Running the notebook I got as far as the LearningSpeedup example
> 
> I've now tried this on three different machines and I can't replicate it... it runs fine for me. :( I'm not sure how to try to debug that....

I've managed to find a machine that replicates this!  The one on the intel loihi cloud gives the same error you are getting!  But only for test('LearningSpeedup', dict(dt=0.0001, learn_rate=1e-4)) -- it doesn't happen with learn_rate=1e-5 (I think -- it's still running, in any case)

tcstewar:
> nengo.exceptions.BuildError: Could not find appropriate wgtExp

I'm pretty sure I now know what that error is, @drasmuss .  The benchmark was using `lambda x: np.random.random()` as the initial function for the learning connection.  This is very hard for LstsqL2 to solve, and I believe the result is slightly sensitive to what version of numpy you are running.  It will generally give weights all near zero, and that can make the `wgtExp` hard to sort out (a variant of #99 ).  So, I've modified the benchmark to allow you to set an initial value for the learned function.  This makes the problem go away, I think.

tcstewar:
I've got all the benchmarks in a reasonable state in terms of results, and they've all been run on Loihi (using their cloud).  Now my plan is to go through and add better labels on the graphs so they're a bit more informative.....

tcstewar:
I've rerun all these benchmarks with the current version of the `remove-passthrough` branch, and that all works fine.  I'm tempted to rebase this branch onto that one, just so we have the same set of commits (this branch fixes the `tau_s` thing in a slightly different way than what go into master).

tbekolay:
Fixed up the history, then moved this elsewhere for internal use (this isn't designed to be part of a release).

Issue #115, opened 2018-10-04
Radius has a different interpretation


arvoelke:
(Preamble: I'm looking at this from the perspective of your average user, and in the context of `Node -> Ensemble` connections. I haven't checked `Ensemble -> Ensemble`, but there's probably something similar due to interneurons.)

In the standard Nengo reference simulator, the radius is post-synapse.

In nengo_loihi, the radius is effectively pre-synapse, and applied individually to each connection.

Another way of thinking of this, is that it's as though each synapse saturates `e.dot(x)` at `[-radius, radius]`.

I understand this has to do with the way the spike generators are set up, but this kind of difference is important to consider, because it changes the way that you have to think about building models, especially when you have multiple connections, and dynamics (higher frequencies). Here's two simple networks to illustrate why this is important to consider:

```python
import numpy as np
import matplotlib.pyplot as plt

import nengo
import nengo_loihi; nengo_loihi.set_defaults()

radius = 1  # set this to 3 to see the difference go away

with nengo.Network(label="Two Minus One") as two_minus_one:
    a = nengo.Node(2)
    b = nengo.Node(-1)
    
    x = nengo.Ensemble(100, 1, radius=radius)
    
    nengo.Connection(a, x)
    nengo.Connection(b, x)
    
    p_x = nengo.Probe(x, synapse=0.1)

synapse = nengo.Lowpass(0.1)
freq = 5

with nengo.Network(label="High-Frequency Input") as high_freq:    
    u = nengo.Node(output=lambda t: np.sin(2*np.pi*freq*t) /
                                    np.abs(synapse.evaluate(freq)))
    y = nengo.Ensemble(100, 1, radius=radius)
    
    nengo.Connection(u, y, synapse=synapse)
    
    p_y = nengo.Probe(y, synapse=0.1)

def compare_simulators(model, p, sim_t):
    plt.figure()
    plt.title(model.label)
    for simulator in (nengo, nengo_loihi):
        with simulator.Simulator(model) as sim:
            sim.run(sim_t)
        plt.plot(sim.trange(), sim.data[p], label=simulator.__name__)
    plt.legend()
    plt.show()

compare_simulators(two_minus_one, p_x, sim_t=10)

compare_simulators(high_freq, p_y, sim_t=1. / freq)
```

The first example implements `2 - 1`. The answer we get should be `1`, but the radius is applied to the `2` to get `1 - 1 = 0` instead.

[...]

PR #124, opened 2018-10-15
Integrator accuracy

### LINKED ISSUE #114: Integrator performs much better with a recurrent transform of 1.12 - 1.15
integrator

I did a little benchmarking of an integrator network, for several different inputs:
 - Cosine
 - Step-function
 - Zero (holding steady at 1)

I measured the RMSE across multiple trials with different seeds, while varying the gain on the recurrent feedback.

For the reference simulator, a recurrent transform of 1 is optimal. But here, I'm getting 1.12 - 1.15, depending on the kind of input. I'm not using any interneurons, and I'm using a fairly long time-constant (100 ms).

Might be useful to eventually have a collection of similar benchmarks in a regression test framework.

```python

from collections import defaultdict

import numpy as np
from pandas import DataFrame

import matplotlib.pyplot as plt
import seaborn as sns

import nengo_loihi; nengo_loihi.set_defaults()
import nengo
from nengo.utils.numpy import rmse
from nengolib.signal import s

n_neurons = 300
conn_tau = 0.1
probe_tau = 0.005
dt = 0.001
sim_t = 2*np.pi
simulator = nengo_loihi

data = defaultdict(list)
for desc, output in [('Cosine', np.cos),
                     ('Step', 1. / sim_t),
                     ('Zero', lambda t: 1 if t <= 1 else 0)]:
    for seed in range(5):
        for transform in np.linspace(1.05, 1.2, 10):
            print("desc=%s, seed=%d, transform=%s" % (desc, seed, transform))

                u = nengo.Node(output=output)
                x = nengo.Ensemble(n_neurons, 1, seed=seed)

                nengo.Connection(u, x, synapse=conn_tau, transform=conn_tau)
                nengo.Connection(x, x, synapse=conn_tau, transform=transform,

                p_x = nengo.Probe(x, synapse=probe_tau)
                p_ideal = nengo.Probe(u, synapse=~s * nengo.Lowpass(probe_tau))

            builder_model = simulator.builder.Model(dt=dt)
            # builder_model.inter_tau = inter_tau
            # assert builder_model.inter_n == 10  # sanity check
            # builder_model.inter_n = inter_n
            # builder_model.inter_noise_exp = 5
            with simulator.Simulator(network=model, model=builder_model, dt=dt) as sim:
                sim.run(sim_t)

            data['x'].append(sim.data[p_x].squeeze())
            data['Input'].append(desc)
            data['RMSE'].append(rmse(sim.data[p_x], sim.data[p_ideal]))
            data['Seed'].append(seed)
            data['Transform'].append(transform)

df = DataFrame(data)
plt.figure()
plt.title("Integration Performance")
sns.lineplot(data=df, x="Transform", y="RMSE", hue="Input", ci=95)
plt.show()
```

Also see #90 which may be related.

hunse:
Improving the accuracy of an integrator network. Fixes #114, namely an integrator using a recurrent weight connection.

The integrator network still does not work as expected when using a decoded (non-weight) connection (now documented in #125).

### DIFF
--- a/nengo_loihi/loihi_api.py
+++ b/nengo_loihi/loihi_api.py
@@ -600,11 +600,15 @@ class SynapseFmt(Profile):
 
         w = np.round(w / 2.**s).clip(-m, m).astype(dtype)
         s2 = s + self.wgtExp
-        shift(w, s2, out=w)
-        np.left_shift(w, 6, out=w)
-
         if s2 < 0:
             warnings.warn("Lost %d extra bits in weight rounding" % (-s2,))
+
+            # round before `s2` right shift, since just shifting would floor
+            # everything resulting in weights biased towards being smaller
+            w = (np.round(w * 2.**s2) / 2**s2).clip(-m, m).astype(dtype)
+
+        shift(w, s2, out=w)
+        np.left_shift(w, 6, out=w)
 
         ws = w // self.scale
         assert np.all(ws <= 255) and np.all(ws >= -256)
--- a/nengo_loihi/loihi_api.py
+++ b/nengo_loihi/loihi_api.py
@@ -99,6 +99,57 @@ def tracing_mag_int_frac(synapses):
     return mag_int, mag_frac
 
 
+def decay_int(x, decay, bits=12, offset=0, out=None):
+    """Decay integer values using a decay constant.
+
+    The decayed value is given by::
+
+        sign(x) * floor(abs(x) * (2**bits - offset - decay) / 2**bits)
+    """
+    if out is None:
+        out = np.zeros_like(x)
+    r = (2**bits - offset - np.asarray(decay)).astype(np.int64)
+    np.right_shift(np.abs(x) * r, bits, out=out)
+    return np.sign(x) * out
+
+
+def decay_magnitude(decay, x0=2**21, bits=12, offset=0):
+    """Estimate the sum of the series of rounded integer decays of ``x0``.
+
+    This can be used to estimate the total input current or voltage (summed
+    over time) caused by an input of magnitude ``x0``. In real values, this is
+    easy to calculate as the integral of an exponential. In integer values,
+    we need to account for the rounding down that happens each time the decay
+    is computed.
+
+    Specifically, we estimate the sum of the series::
+
+        x_i = floor(r x_{i-1})
+
+    where ``r = (2**bits - offset - decay)``.
+
+    To simulate the effects of rounding in decay, we subtract an expected loss
+    due to rounding (``q``) each iteration. Our estimated series is therefore::
+
+        y_i = r * y_{i-1} - q
+            = r^i * x_0 - sum_k^{i-1} q * r^k
+    """
+    # q: Expected loss per time step (found by empirical simulations). If the
+    # value being rounded down were uniformly distributed between 0 and 1, this
+    # should be 0.5 exactly, but empirically this does not appear to be the
+    # case and this value is better (see `test_decay_magnitude`).
+    q = 0.494
+
+    r = (2**bits - offset - np.asarray(decay)) / 2**bits  # decay ratio
+    n = -np.log1p(x0 * (1 - r) / q) / np.log(r)  # solve y_n = 0 for n
+
+    # principal_sum = (1./x0) sum_i^n x0 * r^i
+    # loss_sum = (1./x0) sum_i^n sum_k^{i-1} q * r^k
+    principal_sum = (1 - r**(n + 1)) / (1 - r)
+    loss_sum = q / ((1 - r) * x0) * (n + 1 - (1 - r**(n+1))/(1 - r))
+    return principal_sum - loss_sum
+
+
 def shift(x, s, **kwargs):
     if s < 0:
         return np.right_shift(x, -s, **kwargs)
--- a/nengo_loihi/loihi_cx.py
+++ b/nengo_loihi/loihi_cx.py
@@ -12,6 +12,8 @@ from nengo.utils.compat import is_integer, is_iterable
 from nengo_loihi.loihi_api import (
     BIAS_MAX,
     bias_to_manexp,
+    decay_int,
+    decay_magnitude,
     overflow_signed,
     SynapseFmt,
     tracing_mag_int_frac,
@@ -229,12 +231,16 @@ class CxGroup(object):
             target[:] = new
 
         # --- discretize decayU and decayV
-        u_infactor = (
-            self.decayU.copy() if self.scaleU else np.ones_like(self.decayU))
-        v_infactor = (
-            self.decayV.copy() if self.scaleV else np.ones_like(self.decayV))
         discretize(self.decayU, self.decayU * (2**12 - 1))
         discretize(self.decayV, self.decayV * (2**12 - 1))
+
+        # Compute factors for current and voltage decay. These factors
+        # counteract the fact that for longer decays, the current (or voltage)
+        # created by a single spike has a larger integral.
+        u_infactor = (1. / decay_magnitude(self.decayU, x0=2**21, offset=1)
+                      if self.scaleU else np.ones(self.decayU.shape))
+        v_infactor = (1. / decay_magnitude(self.decayV, x0=2**21)
+                      if self.scaleV else np.ones(self.decayV.shape))
         self.scaleU = False
         self.scaleV = False
 
@@ -824,27 +830,21 @@ class CxSimulator(object):
             group.decayV if group.scaleV else np.ones_like(group.decayV)
             for group in self.groups])
 
-        def decay_float(x, u, d, s):
-            return (1 - d)*x + s*u
-
-        def decay_int(x, u, d, s, a=12, b=0):
-            r = (2**a - b - np.asarray(d)).astype(np.int64)
-            x = np.sign(x) * np.right_shift(np.abs(x) * r, a)  # round to zero
-            return x + u  # no scaling on u
-
         if group_dtype == np.int32:
             assert (self.scaleU == 1).all()
             assert (self.scaleV == 1).all()
-            self.decayU_fn = lambda x, u: decay_int(
-                x, u, d=self.decayU, s=self.scaleU, b=1)
-            self.decayV_fn = lambda x, u: decay_int(
-                x, u, d=self.decayV, s=self.scaleV)
+            self.decayU_fn = (
+                lambda x, u: decay_int(x, self.decayU, offset=1) + u)
+            self.decayV_fn = lambda x, u: decay_int(x, self.decayV) + u
 
             def overflow(x, bits, name=None):
                 _, o = overflow_signed(x, bits=bits, out=x)
                 if np.any(o):
                     self.error("Overflow" + (" in %s" % name if name else ""))
         elif group_dtype == np.float32:
+            def decay_float(x, u, d, s):
+                return (1 - d)*x + s*u
+
             self.decayU_fn = lambda x, u: decay_float(
                 x, u, d=self.decayU, s=self.scaleU)
             self.decayV_fn = lambda x, u: decay_float(
--- a/nengo_loihi/loihi_cx.py
+++ b/nengo_loihi/loihi_cx.py
@@ -231,7 +231,9 @@ class CxGroup(object):
             target[:] = new
 
         # --- discretize decayU and decayV
-        discretize(self.decayU, self.decayU * (2**12 - 1))
+        # subtract 1 from decayU here because it gets added back by the chip
+        decayU = self.decayU * (2**12 - 1) - 1
+        discretize(self.decayU, np.clip(decayU, 0, 2**12 - 1))
         discretize(self.decayV, self.decayV * (2**12 - 1))
 
         # Compute factors for current and voltage decay. These factors
--- a/nengo_loihi/tests/test_loihi_api.py
+++ b/nengo_loihi/tests/test_loihi_api.py
@@ -1,7 +1,9 @@
+from nengo.utils.numpy import rms
 import numpy as np
 import pytest
 
 from nengo_loihi.loihi_api import overflow_signed
+from nengo_loihi.loihi_api import decay_int, decay_magnitude
 
 
 @pytest.mark.parametrize("b", (8, 16, 17, 23))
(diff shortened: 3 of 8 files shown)


hunse:
[integration.pdf](https://github.com/nengo/nengo-loihi/files/2479265/integration.pdf)

Here's the test from #114 run on this branch.

hunse:
Two things: Why are you using `intercepts=nengo.dists.Uniform(-X, X)`? Didn't we decide that `intercepts=nengo.dists.Uniform(-1, X)` is almost always better (i.e. it's the high intercepts that cause problems, not the low ones)? I'd be interested to see those plots redone with that change.

As for the failing test, I'm not seeing that, and neither is Travis CI (I ran it a number of times, too). And it's not a chip thing, because that test doesn't run on the chip (it's just a unit test of a helper function). So maybe something about your Numpy is different, or there's some other difference on your machine? Though looking at your printout, it looks WAY off, so there must be a bug somewhere. I don't get why I'm not seeing it, though.

tcstewar:
> Why are you using `intercepts=nengo.dists.Uniform(-X, X)`? Didn't we decide that `intercepts=nengo.dists.Uniform(-1, X)` is almost always better (i.e. it's the high intercepts that cause problems, not the low ones)? I'd be interested to see those plots redone with that change.

Good point, but the test that you're running (from #114 ) is using `intercepts=nengo.dists.Uniform(-0.5, 0.5)`, so I wanted to make sure I tested that same point.  I'll try the (-1, X) too, just in case.  :)

> As for the failing test, I'm not seeing that, and neither is Travis CI (I ran it a number of times, too). And it's not a chip thing, because that test doesn't run on the chip (it's just a unit test of a helper function). So maybe something about your Numpy is different, or there's some other difference on your machine? Though looking at your printout, it looks WAY off, so there must be a bug somewhere. I don't get why I'm not seeing it, though.

Interesting....  I just tried it on the intel cloud and it also passes there.   `pip freeze` claims that on both the intel cloud (where it passes) and on my laptop (where it fails), I'm on `numpy==1.14.3`.  I am on Windows on my laptop, but that really shouldn't matter....  I'm on Python 3.6.1 on the laptop and the intel cloud is Python 3.5.2...

hunse:
One thing I'm puzzled about is why there are such significant differences between the emulator and the chip. With `max_rates=120` they look pretty similar (though even there, you can see a bit of evidence of the chip being noisier, e.g. with `intercepts=0.5` and `transform=1.05`). But with higher firing rates, they're pretty different: On the emulator, `transform=1.0` seems pretty clearly to be the best, whereas on the chip it looks like a slightly smaller transform (0.95) would be better. It's possible this has to do with noisy interneurons; I'd be curious to try this with #132 to see if that makes any difference. I assume that if you're using `precompute=False`, you've set the number of snip spikes high enough that nothing is getting dropped?

Finally, the biggest thing that concerns me is that performance deteriorates a lot for higher numbers of neurons, both in emulator and in chip, before and after these changes. So that's obviously something for future work, but that seems like pretty important future work to me.

tcstewar:
> One thing I'm puzzled about is why there are such significant differences between the emulator and the chip. With `max_rates=120` they look pretty similar (though even there, you can see a bit of evidence of the chip being noisier, e.g. with `intercepts=0.5` and `transform=1.05`). But with higher firing rates, they're pretty different: On the emulator, `transform=1.0` seems pretty clearly to be the best, whereas on the chip it looks like a slightly smaller transform (0.95) would be better.

One possibility is that it might be the same weirdness that's giving me the failing test.  I'll re-run the simulator tests on the intel cloud and see what that looks like.

> It's possible this has to do with noisy interneurons; I'd be curious to try this with #132 to see if that makes any difference.

I don't think there are interneurons in this model: the solver has `weights=True`

> I assume that if you're using `precompute=False`, you've set the number of snip spikes high enough that nothing is getting dropped?

This is with `precompute=True`, for speed reasons (I was running a *lot* of sims...)....  I can try without precompute too....

drasmuss:
`test_decay_magnitude` also fails for me:
```
>       assert np.all(relative_diff < 1e-6)
E       assert False
E        +  where False = <function all at 0x0000014677028730>(array([ 9.71177396e-01, -1.08811599e-03, -6.13906343e-04, -4.32027241e-04,\n       -3.39373637e-04, -2.80564914e-04, -2...7, -3.82642914e-07,\n       -9.86391045e-08, -7.17645022e-07, -1.80609154e-07, -4.31690146e-07,\n       -1.07072338e-07]) < 1e-06)
E        +    where <function all at 0x0000014677028730> = np.all
```
That's on `numpy==1.14.5`

tcstewar:
> `test_decay_magnitude` also fails for me:

I found the cause!  Here's the line in the test:

```python
s = np.sum(ys, axis=0)
```

that needs to be changed to

```python
s = np.sum(ys, axis=0, dtype=np.int64)
```

(the sum is silently overflowing as on some machines it seems to default to using int32)

arvoelke:
> (the sum is silently overflowing as on some machines it seems to default to using int32)

FYI, https://github.com/numpy/numpy/issues/9464 -- Doesn't sound like there's any way to force int64 to be the default at the moment. :/

hunse:
Well that's very good to know about that difference. I had assumed that `np.array([1])` would always return an array of int64; I didn't realize it was machine-specific.

@tcstewar @drasmuss: can you check out 2eef211 (the new version of this branch) and make sure that fixes the test for you?

drasmuss:
Test passes for me :+1:

hunse:
> For what it's worth, this branch seems to dramatically improve (up to 2x improvement in accuracy versus master) the Delay Network for lower recurrent taus (e.g., 10ms). For larger recurrent taus (e.g., 100ms)---with the same readout tau as before---they both seem to perform the same.

Looking at those little top-right corner plots (that's what we're talking about here, right?), it appears to me that it's the opposite. I assume the x-axis is the tau you're talking about here. It looks like for up to taus of around 50 ms (0.05 s), the plots are roughly the same. For example, at 0.05 s, they both appear to be around 0.15 NRMSE. It's for larger taus that there's a difference, for example at 0.1 s you see this 2x difference you were referring to (with the old one being around 0.3 NRMSE, and the new one around 0.15).

Or is the tau on the x-axis of those little plots the "readout tau", which I assume is the delay that the network is trying to mimic or read out? And then you did analyses like the one above for many different recurrent taus, but you're just showing us the analysis for a shorter one (around 5 to 10 ms)?

arvoelke:
The x-axis is the time point within the window. Both plots are generated by a single trial with a fixed tau. The state of the DN decodes the entire window, and then we're plotting the accuracy (y-axis) of that window (x-axis).

hunse:
So the results that you're talking about---that this PR improves things for shorter recurrent taus---is not demonstrated by the plots above, since those are just for a single recurrent tau. Is that correct?

When you look at the error for a particular recurrent tau, is that the integrated error over that whole window you're talking about (i.e. for any possible readout delay in the range 0 to 0.1 s)? Or is it just the error at the longest readout delay (0.1 s)?

arvoelke:
> So the results that you're talking about---that this PR improves things for shorter recurrent taus---is not demonstrated by the plots above, since those are just for a single recurrent tau. Is that correct?
>
> When you look at the error for a particular recurrent tau, is that the integrated error over that whole window you're talking about (i.e. for any possible readout delay in the range 0 to 0.1 s)? 

That's right. The main thing I was pointing out is that the overall error (y-axis) is better by a factor of ~2x for this branch versus master.

We can talk about the longer taus but I think something much stranger is happening that could be somewhat orthogonal. I need to look into this. It could have to do with the spike generator on the input side in conjunction with the input being scaled by tau.

arvoelke:
This is what it looks like for the longer taus (on either branch):

loihi_delay_network

And this is what it's supposed to look like (simply changing `nengo_loihi.Simulator` to `nengo.Simulator`):

nengo_delay_network

Again I think it has something to do with the spike generator. Maybe https://github.com/nengo/nengo-loihi/issues/115 as well. Any hints/ideas would also be helpful! In either case I think this is mostly independent of the improvements of this branch.

arvoelke:
And FWIW again here is an integrator running on this branch:

integrator-accuracy

versus master:

integrator-master

Both are averaged across 200 trials, use the ReLU model, and employ a number of "tricks" (see https://github.com/nengo/nengo-loihi/issues/115#issuecomment-447015272). Would be nice to have this in master so that my thesis doesn't need to reference a branch. :+1: (Code available upon request.) Willing to review if that helps.

nengo_loihi/loihi_api.py
```diff
@@ -99,6 +99,46 @@ def tracing_mag_int_frac(synapses):
     return mag_int, mag_frac
 
 
+def decay_int(x, decay, bits=12, offset=0, out=None):
+    """Decay integer values using a decay constant.
+
+    The decayed value is given by
+
+        sign(x) * floor(abs(x) * (2**bits - offset - decay) / 2**bits)
```
tbekolay: This will render as a blockquote, which is probably not what you want. If you want it to show up like math, use the `.. math::` directive; if you want it to look like code you can either end the previous line with `::` or use the `.. code-block::` directive; see https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html

nengo_loihi/loihi_api.py
```diff
@@ -99,6 +99,46 @@ def tracing_mag_int_frac(synapses):
...
+    """
+    if out is None:
+        out = np.zeros_like(x)
+    r = (2**bits - offset - np.asarray(decay)).astype(np.int64)
+    np.right_shift(np.abs(x) * r, bits, out=out)
+    return np.sign(x) * out
+
+
+def decay_magnitude(decay, x0=2**21, bits=12, offset=0):
+    """Estimate the sum of the series of rounded integer decays of `x0`.
```
tbekolay: Single backticks here will likely give a Sphinx warning. If you want it to render like code, use double backticks. Some people use italics (`*x0*`) for parameter arguments instead; we haven't standardized so whatever you prefer.

nengo_loihi/loihi_api.py
```diff
@@ -99,6 +99,46 @@ def tracing_mag_int_frac(synapses):
...
+
+    This can be used to estimate the total input current or voltage (summed
+    over time) caused by an input of magnitude `x0`.
```
tbekolay: Single backticks here too

nengo_loihi/loihi_api.py
```diff
@@ -99,6 +99,46 @@ def tracing_mag_int_frac(synapses):
...
+
+    Specifically, we estimate the sum of the series
+        x_i = floor(r * x_{i-1})
```
tbekolay: Need `math` or `code-block` directive.

nengo_loihi/loihi_api.py
```diff
@@ -99,6 +99,46 @@ def tracing_mag_int_frac(synapses):
...
+    where ``r = (2**bits - offset - decay)``.
+
+    To simulate the effects of rounding in decay, we subtract an expected loss
+    due to rounding (`q`) at each iteration. Our estimated series is therefore:
+        y_i = r * y_{i-1} - q
```
tbekolay: `math` or `code-block`

nengo_loihi/loihi_api.py
```diff
@@ -99,6 +99,46 @@ def tracing_mag_int_frac(synapses):
...
+
+    This can be used to estimate the total input current or voltage (summed
+    over time) caused by an input of magnitude `x0`.
+
+    Specifically, we estimate the sum of the series
+        x_i = floor(r * x_{i-1})
+    where ``r = (2**bits - offset - decay)``.
+
+    To simulate the effects of rounding in decay, we subtract an expected loss
+    due to rounding (`q`) at each iteration. Our estimated series is therefore:
```
tbekolay: Single backticks

nengo_loihi/loihi_cx.py
```diff
@@ -13,6 +13,8 @@
     BIAS_MAX,
     bias_to_manexp,
     overflow_signed,
+    decay_int,
```
tbekolay: These should be in alphabetic order, so before `overflow_signed`

nengo_loihi/tests/test_loihi_api.py
```diff
@@ -1,7 +1,10 @@
 import numpy as np
 import pytest
 
+from nengo.utils.numpy import rms
```
tbekolay: From the perspective of Nengo Loihi, Nengo is a third-party dependency like all the rest, so it should go in the upper block with numpy and pytest.

nengo_loihi/tests/test_loihi_api.py
```diff
@@ -1,7 +1,10 @@
...
+
 from nengo_loihi.loihi_api import overflow_signed
+from nengo_loihi.loihi_api import decay_magnitude, decay_int
```
tbekolay: Alphabetical so `decay_int` should go before `decay_magnitude`.

nengo_loihi/tests/test_loihi_api.py
```diff
@@ -22,3 +25,71 @@ def test_overflow_signed(b):
     y, o = overflow_signed(x, bits=b)
     assert np.array_equal(y, z)
     assert np.array_equal(o, q)
+
+
+@pytest.mark.parametrize('offset', (0, 1))
+def test_decay_magnitude(offset, plt, logger):
+    bits = 12
+    # decays = np.arange(1, 2**bits)
```
tbekolay: Probably remove this? Guessing it was changed because it took too long?

nengo_loihi/tests/test_loihi_api.py
```diff
@@ -22,3 +25,71 @@ def test_overflow_signed(b):
...
+    decays = np.arange(1, 2**bits, 7)
+    ref = []
+    emp = []
+    est = []
+
+    for decay in decays:
+        def empirical_decay_magnitude(decay, x0, invert=False):
```
tbekolay: The `invert` arg is never used, and it feels like a weird thing to make part of the function when you could just do `1.0 / empirical_decay_magnitude(...)` instead. I'd remove it.

nengo_loihi/tests/test_loihi_api.py
```diff
@@ -22,3 +25,71 @@ def test_overflow_signed(b):
...
+                    break
+                y = decay_int(y, decay, bits=bits, offset=offset)
+            else:
+                raise RuntimeError("Exceeded max number of iterations")
+
+            return x0 / y_sum if invert else y_sum / x0
+
+        # x0 = np.arange(2**10, 2**11, dtype=np.int32)
+        x0 = np.arange(2**21 - 1000, 2**21, step=41, dtype=np.int32)
+        # x0 = np.arange(2**21 - 1000, 2**21, dtype=np.int32)
```
tbekolay: This and two lines above should be removed? Or parametrize the test if we should test with all three.

nengo_loihi/tests/test_loihi_api.py
```diff
@@ -22,3 +25,71 @@ def test_overflow_signed(b):
...
+
+        m = empirical_decay_magnitude(decay * np.ones_like(x0), x0)
+        m0 = m.mean()
+        emp.append(m0)
+
+        # reference (naive) method, not accounting for truncation loss
+        r = (2**bits - offset - decay) / 2**bits
+        # Sx0 = 1. / (1 - r)  # sum_i^n x_i/x_0, where x_n = r**n*x_0 = 0
+        # ref.append(Sx0)
```
tbekolay: This and the line above should be removed?

nengo_loihi/loihi_api.py
hunse: Can we talk about standards for how we do equations? I find LaTeX great for stuff that will definitely be rendered, but these docstrings may either be rendered (for online) or just as strings (when reading the code, or calling `help` on a function).

For that reason, I think that we should stay away from the `math` directive, unless we have a really good reason for it in a particular situation. I'd opt for using `::` and displaying stuff as code wherever possible, since this is clear in either raw or rendered format.

hunse: Having said that, I realize that some things are hard to do in code (or at least need a standard way of writing them), for example a subscript like `x_{i-1}`. Sums are another big one that I tend to write in LaTeX notation. So I'm not sure which way to go on this.

nengo_loihi/loihi_api.py
hunse: done

nengo_loihi/tests/test_loihi_api.py
hunse: Yeah. I kind of left it in as a reminder, that if you're making big changes to `decay_magnitude`, then you might want to uncomment this and run the longer test. But it's not a test that we need to run all the time. I don't think we have a good way to handle tests like this right now (where we want a shorter and longer version).

hunse:
Comments addressed.

hunse:
I realized I forgot to address your original (not-inline) comment, so I added a few commits to do that too. Between the comment briefly describing `u_factor` and `v_factor` when they're declared, and the addition to the `decay_magnitude` docstring, I think we've got all the key points of my comment at: https://github.com/nengo/nengo-loihi/issues/114#issuecomment-429103026

The one thing I don't talk about is how to choose the initial value for computing the decay magnitude (`x0`). I didn't really have a principled way to do this, so there's not really much to say. The most I can think of is a note somewhere that this is a thing and we could look into it more at some point. But a) I don't think it's particularly significant, b) there are lots of things to look into in the code, and c) if this does need revisiting, I expect it will be because of some problem (a particular model is not as accurate as we'd like) and that will inspire us to revisit it.

Issue #73, opened 2018-09-19
Auto-setting defaults


tbekolay:
The nengo tests and benchmarks would likely see improvements if we automatically call `nengo_loihi.set_defaults` before creating models. Should we do this? It is straightforward to do this in pytest, and I presume we can do something similar with benchmarking. But it could be considered magical when the model is either knowingly or unknowingly relying on the Nengo default.

Note that this is a separate issue from #69; the intercepts should be limited regardless of what the default or user setting is for intercepts. With `set_defaults`, we only change the default values, so any values specified by the user will be used as is.

tcstewar:
> The nengo tests and benchmarks would likely see improvements if we automatically call `nengo_loihi.set_defaults` before creating models.

I'm not entirely sure that's true any more.  #69 greatly improves things when using the normal defaults, and `nengo.loihi.set_defaults` is pretty aggressive with its intercept setting (-0.5 to 0.5), and I'm not sure we've explored how much that affects things.  And #49 made things much better for high-firing neurons, so we might not need to reduce the firing rates to (100, 120).  In fact, reducing the firing rates that much might also cause tests and benchmarks to get worse.  

I'll try running a few benchmarks with and without set_defaults and see what the results look like....  :)

hunse:
Yeah, let's definitely explore the effects of setting vs not setting the defaults. We can also change the defaults if we want. (My choices of -0.5 to 0.5 for the intercepts and 100 to 120 for the max rates were both pretty arbitrary. They were values that seemed to work well for early models, but a lot has changed since then, and I didn't do any parameter sweeps or anything.)

With regards to setting the defaults automatically, though, I'm not sure how we would implement this. When we're creating the model, we don't know what simulator it's going to be run on. We could see if `nengo_loihi` is imported, but then that'll give different results if the user imports it at the top (proper) vs when they create the simulator (which I sometimes do when hacking). And once the model is created, I don't think we have any way to know what was a default versus what was not (other than checking if the set value equals the default, but if we did that, then there would be no way for the user to say e.g. "I really want the intercepts to be Uniform(-1, 1), even on nengo_loihi").

PR #126, opened 2018-10-15
Configure defaults for optimal Loihi performance

hunse:
In #73, we started talking about whether the current defaults are actually the best. I chose them pretty arbitrarily, so it's unlikely that they are.

This PR is a place for us to collect changes to the defaults that improve performance across a variety of models. With many of the Nengo benchmarks now becoming available, those could be a good way to test potential changes. 

To start, I'm correcting something that was a mistake on my part. Since intercepts happen after encoders, we actually just need to lower the upper bound on intercepts (away from one) to avoid high gains. The lower bound can (and I think should) stay at -1. I'd like to test this change on some benchmarks, though, to see if it's actually better.

EDIT: I just went and rebased this to the `integrator-accuracy` branch (#124), since that makes some pretty significant changes to how we choose weights that could have an effect on accuracy measurements.

### DIFF
--- a/.nengobones.yml
+++ b/.nengobones.yml
@@ -409,6 +409,10 @@ setup_cfg:
       test_solvers.py:test_non_compositional_solver_transform_error:
         1D convolution not supported
 
+      # sparse transforms not supported
+      test_transforms.py:test_sparse[*:
+        sparse transforms not supported
+
 
 docs_conf_py:
   intersphinx_mapping:
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -65,6 +65,9 @@ Release history
 - We no longer disable the Nengo decoder cache for all models.
   (`#202 <https://github.com/nengo/nengo-loihi/pull/202>`__,
   `#207 <https://github.com/nengo/nengo-loihi/issues/207>`__)
+- Transforms to on-chip neurons are now applied on-chip,
+  which avoids scaling issues and large off-chip transforms.
+  (`#126 <https://github.com/nengo/nengo-loihi/pull/126>`__)
 
 0.6.0 (February 22, 2019)
 =========================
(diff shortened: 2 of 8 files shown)


hunse:
This is also related to #83, since how we scale weights will determine how sensitive we are to large weights/gains, which is the reason for limiting intercepts.

tbekolay:
After rebasing, the only commit not yet incorporated in master is bd5df976afdfe4acc542a3bd1e609dbd1b22bdef which I believe is an improvement over the current state of master.

drasmuss:
This causes one of the nengo core tests to fail (see https://travis-ci.com/nengo/nengo-loihi/jobs/205047436#L3674).

hunse:
Hmm, that's strange. That test is making sure that inhibition works correctly to silence neurons.

In core Nengo, you get the output you'd expect. The bottom plot is showing neuron voltages.
[test_connection.test_node_to_neurons.pdf](https://github.com/nengo/nengo-loihi/files/3249134/test_connection.test_node_to_neurons.pdf)

Even on master of this repository, surprisingly some neurons are firing, however not enough to break the tolerances and fail the test.
[test_connection.test_node_to_neurons.pdf](https://github.com/nengo/nengo-loihi/files/3249138/test_connection.test_node_to_neurons.pdf)

This branch seems to make things worse, with more neurons firing when they should be silent.
[test_connection.test_node_to_neurons.pdf](https://github.com/nengo/nengo-loihi/files/3249143/test_connection.test_node_to_neurons.pdf)

So something is definitely going on here that shouldn't be, but the problem seems to go beyond this branch.

hunse:
Figured it out. When we do node->neuron connections, we do the transform on the host (before we turn the value into spikes). This works fine for transforms that keep the output in the range `[-1, 1]`, but for ones like in this test (the input is 1 and the transform is -5), it doesn't work so well. Not sure what the best fix is.

hunse:
Ok, we're now putting weights on the chip when doing any connection into neurons. I think this has a number of advantages (see the commit message 32a5873). It does make the mapping behaviour slightly more complicated, since now there are differences depending on both the pre and post objects.

PR #128, opened 2018-10-18
Call finishRun before disconnect

tbekolay:
Currently when we finish a simulation, we call `N2Board.disconnect`. Apparently there is also a `finishRun` method that we have not been calling, but it seems that when using snips, it's important to call this method.


hunse:
So the only thing I'm not sure about is if it's bad to call `finishRun` twice. In normal behaviour (no exception) we call it in `simulator.py`, it's just if there's an exception that it doesn't get called. If this branch runs fine on the chip, I would assume it's fine to call it twice.

tbekolay:
Closing this as we are calling `finishRun` at an appropriate place.

PR #131, opened 2018-10-22
Fix conftest Simulator target for Nengo tests

hunse:
When we run the Nengo tests (with `--pyargs nengo`), the `target` option was being ignored because Nengo's `Simulator` fixture is being used, rather than the one in `nengo_loihi`. This sets things up so that we'll keep the proper target for Nengo tests too.


hunse:
This is fixed in `nengo_tests`, just didn't realize.

Issue #129, opened 2018-10-18
Addition of two inputs is not very accurate for some inputs


tcstewar:
When we have multiple inputs to an Ensemble, we should get addition.  However, for some values this doesn't seem to be very accurate.  For example:

```python
model = nengo.Network(seed=1)
with model:
    stim_a = nengo.Node(0.5)
    stim_b = nengo.Node(0.5)
    
    a = nengo.Ensemble(n_neurons=100, dimensions=1)
    b = nengo.Ensemble(n_neurons=100, dimensions=1)
    
    nengo.Connection(stim_a, a)
    nengo.Connection(stim_b, b)
    
    c = nengo.Ensemble(n_neurons=100, dimensions=1)    
    nengo.Connection(a, c)
    nengo.Connection(b, c)
    
    p_c = nengo.Probe(c, synapse=0.1)
    p_a = nengo.Probe(a, synapse=0.1)    
    p_b = nengo.Probe(b, synapse=0.1)        
```
![image](https://user-images.githubusercontent.com/2380772/47138692-18e75880-d288-11e8-91e8-17009ccb7394.png)

This problem is much reduced by decreasing dt:
```python
sim2 = nengo_loihi.Simulator(model, dt=0.0005)
```
![image](https://user-images.githubusercontent.com/2380772/47138810-6794f280-d288-11e8-9097-2059a3404621.png)

```python
sim2 = nengo_loihi.Simulator(model, dt=0.0001)
```
![image](https://user-images.githubusercontent.com/2380772/47138835-75e30e80-d288-11e8-8df8-5e85c4c8b194.png)

Alternatively, we can make the problem completely go away by switching to a weight solver (and thus not having interneurons):

```python
    solver = nengo.solvers.LstsqL2(weights=True)
    nengo.Connection(a, c, solver=solver)
    nengo.Connection(b, c, solver=solver)
```
![image](https://user-images.githubusercontent.com/2380772/47138906-9b701800-d288-11e8-9743-735b641d3236.png)

I believe this may have something to do with the integrator stability problems (#114 and the nengo_benchmarks), as the integrator is supposed to work by adding its input to the recurrent value.

Interestingly, some input values work better than others:

```python
    stim_a = nengo.Node(1)
    stim_b = nengo.Node(-1)
```

![image](https://user-images.githubusercontent.com/2380772/47139334-8d6ec700-d289-11e8-8625-b1a9f7d605e1.png)

```python
    stim_a = nengo.Node(0.5)
[...]

PR #132, opened 2018-10-25
Interneuron improvements

hunse:
This helps to address #129.

Rather than using homogeneous interneurons with noise, this uses heterogeneous interneurons. However, they are designed so that they all use the same decoder, which means that the size of encoders in target populations can remain small (otherwise it would scale with the number of interneurons).

Based off #124.

### DIFF
--- a/nengo_loihi/builder.py
+++ b/nengo_loihi/builder.py
@@ -26,9 +26,314 @@ from nengo_loihi.loihi_cx import (
     CxSpikeInput,
     CxSynapses,
 )
-from nengo_loihi.neurons import loihi_rates
+from nengo_loihi.neurons import (
+    loihi_rates,
+    LoihiSpikingRectifiedLinear,
+    NIF,
+)
 
 logger = logging.getLogger(__name__)
+
+
+class DecodeNeurons(object):
+    """Defines parameters for a group of decode neurons.
+
+    DecodeNeurons are used on the chip to facilitate NEF-style connections,
+    where activities from a neural ensemble are first transformed into a
+    decoded value (which is stored in the activities and synapses of the
+    spiking decode neurons), before being passed on to another ensemble
+    (via that ensemble's encoders).
+
+    Parameters
+    ----------
+    dt : float
+        Time step used by the simulator.
+    """
+    def __init__(self, dt=0.001):
+        self.dt = dt
+
+    def __str__(self):
+        return "%s(dt=%0.3g)" % (type(self).__name__, self.dt)
+
+    def get_cx(self, weights, cx_label=None, syn_label=None):
+        """Get a CxGroup for implementing neurons on the chip.
+
+        Parameters
+        ----------
+        weights : (d, n) ndarray
+            Weights that project the ``n`` inputs to the ``d`` dimensions
+            represented by these neurons. Typically, the inputs will be neurons
+            belonging to an Ensemble, and these weights will be decoders.
+        cx_label : string (Default: None)
+            Optional label for the CxGroup.
+        syn_label : string (Default: None)
+            Optional label for the CxSynapses.
+
+        Returns
+        -------
+        cx : CxGroup
+            The neurons on the chip.
+        syn : CxSynapses
+            The synapses connecting into the chip neurons.
+        """
+        raise NotImplementedError()
+
+    def get_ensemble(self, dim):
+        """Get a Nengo Ensemble for implementing neurons on the host.
+
+        Parameters
+        ----------
+        dim : int
+            Number of dimensions to be represented by these neurons.
+
+        Returns
+        -------
+        ens : Ensemble
+            An Ensemble for implementing these neurons in a Nengo network.
+        """
+        raise NotImplementedError()
+
+    def get_post_encoders(self, encoders):
+        """Encoders for post population that these neurons connect in to.
+
+        Parameters
+        ----------
+        encoders : (n, d) ndarray
+            Regular scaled encoders for the ensemble, which map the ensemble's
+            ``d`` input dimensions to its ``n`` neurons.
+
+        Returns
+        -------
+        decode_neuron_encoders : (?, n) ndarray
+            Encoders for mapping these neurons to the post-ensemble's neurons.
+            The number of rows depends on how ``get_post_inds`` is being used
+            (i.e. there could be one row per neuron in this group, or there
+            could be fewer rows with ``get_post_inds`` mapping multiple neurons
+            to each row).
+        """
+        raise NotImplementedError()
+
+    def get_post_inds(self, inds, d):
+        """Indices for mapping neurons to post-encoder dimensions.
+
+        Parameters
+        ----------
+        inds : list of ints
+            Indices for mapping decode neuron dimensions to post-ensemble
+            dimensions. Usually, this will be determined by a slice on the
+            post ensemble in a connection (which maps the output of the
+            transform/function to select dimensions on the post ensemble).
+        d : int
+            Number of dimensions in the post-ensemble.
+        """
+        raise NotImplementedError()
+
+
+class OnOffDecodeNeurons(DecodeNeurons):
+    """One or more pairs of on/off neurons per dimension.
+
+    In this class itself, all the pairs in a dimension are identical. It can
+    still be advantageous to have more than one pair per dimension, though,
+    since this can allow all neurons to have lower firing rates and thus
+    act more linearly (due to period aliasing at high firing rates). Subclasses
+    may use pairs that are not identical (by adding noise or heterogeneity).
+
+    Parameters
+    ----------
+    pairs_per_dim : int
+        Number of repeated neurons per dimension. Currently, all DecodeNeuron
+        classes use separate on/off neuron pairs for each dimension. This is
+        the number of such pairs per dimension.
+    dt : float
+        Time step used by the simulator.
+    rate : float (Default: None)
+        Max firing rate of each neuron. By default, this is chosen so that
+        the sum of all repeated neuron rates is ``1. / dt``, and thus as a
+        group the neurons average one spike per timestep.
+    """
+
+    def __init__(self, pairs_per_dim=1, dt=0.001, rate=None):
+        super(OnOffDecodeNeurons, self).__init__(dt=dt)
+
+        self.pairs_per_dim = pairs_per_dim
+
+        self.rate = (1. / (self.dt * self.pairs_per_dim) if rate is None
+                     else rate)
+        self.scale = 1. / (self.dt * self.rate * self.pairs_per_dim)
+        self.neuron_type = LoihiSpikingRectifiedLinear()
+
+        gain = 0.5 * self.rate * np.ones(self.pairs_per_dim)
+        bias = gain  # intercept of -1
+        self.gain = gain.repeat(2)
+        self.bias = bias.repeat(2)
+        # ^ repeat for on/off neurons
+
+    def __str__(self):
+        return "%s(pairs_per_dim=%d, dt=%0.3g, rate=%0.3g)" % (
+            type(self).__name__, self.pairs_per_dim, self.dt, self.rate)
+
+    def get_cx(self, weights, cx_label=None, syn_label=None):
+        gain = self.gain * self.dt
+        bias = self.bias * self.dt
+
+        d, n = weights.shape
+        n_neurons = 2 * d * self.pairs_per_dim
+        cx = CxGroup(n_neurons, label=cx_label, location='core')
+        cx.configure_relu(dt=self.dt)
+        cx.bias[:] = bias.repeat(d)
+
+        syn = CxSynapses(n, label=syn_label)
+        weights2 = []
+        for ga, gb in gain.reshape(self.pairs_per_dim, 2):
+            weights2.extend([ga*weights.T, -gb*weights.T])
+        weights2 = np.hstack(weights2)
+        syn.set_full_weights(weights2)
+        cx.add_synapses(syn)
+
+        return cx, syn
+
+    def get_ensemble(self, dim):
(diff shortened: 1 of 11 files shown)


tbekolay:
I don't want this to derail this PR, but I did want to note that the first time someone mentioned "interneurons" to me I was somewhat confused because the fact that they are used for NEF decoding wasn't clear in the name, so I assumed that they were inhibitory interneurons as we've used in some other contexts (e.g. respecting Dale's principle). Yes I know that not all biological interneurons are inhibitory, but I also haven't seen the term interneuron used before in the sense of decoding values. Since this PR adds several interneuron-related elements to the API, it might be a good time to explicitly decide if we're ok with the name interneuron or if we can come up with something more clear. A few alternative names come to mind:

- Decode neurons / interneurons
- NEF neurons / interneurons
- Relay neurons / interneurons (this is also a neuroscience term)

My vote would be for decode neurons as it's more clear what their purpose is (though obviously it's never going to be perfectly clear with one word), and isn't a term already used in neuroscience.

tcstewar:
> My vote would be for decode neurons as it's more clear what their purpose is (though obviously it's never going to be perfectly clear with one word), and isn't a term already used in neuroscience.

+1 to that suggestion!

nengo_loihi/builder.py
```diff
@@ -364,16 +585,18 @@ def build_ensemble(model, ens):
         bias=bias)
 
 
-def build_interencoders(model, ens):
-    """Build encoders accepting on/off interneuron input."""
+def build_inter_encoders(model, ens, kind='inter_encoders'):
```
tbekolay: Why not pass in the `Interneurons` instance here rather than a string? With the class hierarchy, it should be fine to pass in any instance.

nengo_loihi/splitter.py
```diff
@@ -289,13 +289,10 @@ def split_host_to_chip(networks, conn):
...
-        2 * dim, dim,
-        neuron_type=NIF(tau_ref=0.0),
-        encoders=np.vstack([np.eye(dim), -np.eye(dim)]),
-        max_rates=np.ones(dim * 2) * networks.max_rate,
-        intercepts=np.ones(dim * 2) * -1,
-        add_to_container=False)
+    if networks.inter_neurons is None:
+        raise BuildError("Interneurons must be specified for host->chip "
+                         "connection.")
+    ens = networks.inter_neurons.get_ensemble(dim)
```
tbekolay: If `get_ensemble` is expected to exist for all `Interneuron` instances, it should be part of the top-level `Interneuron` base class (currently it does not).

nengo_loihi/tests/test_interneurons.py
```diff
@@ -0,0 +1,72 @@
...
+])
+def test_add_inputs(interneurons, tolerance, Simulator, seed, plt):
+    sim_time = 2.
+    pres_time = 0.5
+    eval_time = 0.25
+
+    stim_fn_a = nengo.processes.Piecewise({
+        t: (2*(i % 2) - 1) * 0.4
+        for i, t in enumerate(np.arange(0, sim_time, pres_time))})
+    stim_fn_b = stim_fn_a
```
tbekolay: Why is this a new variable? This could just be `stim_fn` and passed to both nodes?

nengo_loihi/tests/test_interneurons.py
```diff
@@ -0,0 +1,72 @@
...
+    tmask = np.zeros(t.shape, dtype=bool)
+    for pres_t in np.arange(0, sim_time, pres_time):
+        t0 = pres_t + pres_time - eval_time
+        t1 = pres_t + pres_time
+        tmask |= (t >= t0) & (t <= t1)
+
+    target = sim.data[p_stim_a] + sim.data[p_stim_b]
+
+    error = np.abs(sim.data[p_c][tmask] - target[tmask]).mean()
+    print(error)
```
tbekolay: I'm guessing this was here to get good tolerance values initially? I would probably remove it in lieu of maybe something like `plt.title("error=%g" % error)`

nengo_loihi/tests/test_interneurons.py
```diff
@@ -0,0 +1,72 @@
...
+
+        nengo.Connection(stim_a, a)
+        nengo.Connection(stim_b, b)
+
+        c = nengo.Ensemble(n_neurons=100, dimensions=1)
+        nengo.Connection(a, c)
+        nengo.Connection(b, c)
+
+        out_synapse = nengo.Alpha(0.03)
+        stim_synapse = out_synapse.combine(
```
tbekolay: This feels weird. Why is `stim_synapse` like this?

nengo_loihi/builder.py
```diff
@@ -31,6 +31,250 @@
...
+        from nengo_loihi.neurons import LoihiSpikingRectifiedLinear
+        self.neuron_type = LoihiSpikingRectifiedLinear()
+
+        self.dt = dt
+
+        # firing rate of inter neurons
+        self._inter_rate = None
+
+        # number of inter neurons
+        self.n = n
```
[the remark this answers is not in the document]
hunse: So `n` isn't actually the total number of neurons. It's the number of neurons per dimension per on/off pair (at least in all the interneuron types we have so far). This is because all our interneurons use a system of on/off pairs and separate interneurons for each dimension. Theoretically, you could have interneurons where it's just one big group of neurons, and each responds to multiple dimensions (just like an ensemble with multiple dimensions).

So we could call it `neurons_per_dimension`, but even that wouldn't be quite accurate because the actual neurons per dimension would be twice that (because we always have on/off pairs).

nengo_loihi/tests/test_interneurons.py
hunse: Each time there's a connection in the network, there's two `Lowpass(0.005)` filters applied (one from the ensemble and one from the interneurons). Two identical lowpass filters make an alpha filter, and I combine one of them for each connection. This replicates the filtering and delay that happens in the network, so that the stim signal and recorded output signal are comparable.

nengo_loihi/tests/test_interneurons.py
hunse: Partly to get good initial values, but also because this is somewhat a benchmark in addition to a test. I use this test to make changes to different interneuron classes and see how it affects the performance.

My plan was to use the logger fixture once the PR where I add that gets merged, but I still have some problems with the logger (i.e. it would be nice to be able to just print everything that it logs to the screen, with a command line option like `--log-to-screen`).

nengo_loihi/builder.py
```diff
@@ -31,6 +31,250 @@
 logger = logging.getLogger(__name__)
 
 
+class Interneurons(object):
```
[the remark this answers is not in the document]
hunse: Yeah, this all gets split up in refactoring.

nengo_loihi/builder.py
```diff
@@ -31,6 +31,250 @@
...
+    """Defines parameters for a group of interneurons.
+
+    Interneurons are used on the chip to facilitate NEF-style connections,
+    where activities from a neural ensemble are first transformed into a
+    decoded value (which is stored in the activities and synapses of the
+    spiking interneurons), before being passed on to another ensemble
+    (via that ensemble's encoders).
+    """
```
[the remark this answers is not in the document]
hunse: ok

nengo_loihi/builder.py
```diff
@@ -31,6 +31,250 @@
...
+    def get_post_encoders(self, encoders):
+        encoders = encoders * self.inter_scale()
+        return np.vstack([encoders.T, -encoders.T])
+
+    def get_cx(self, weights, cx_label=None, syn_label=None):
+        self.fix_parameters()
+        d, n = weights.shape
+        cx = CxGroup(2 * d * self.n, label=cx_label, location='core')
+        cx.configure_relu(dt=self.dt)
+        cx.bias[:] = self.bias * np.ones(d * 2 * self.n)
```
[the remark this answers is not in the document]
hunse: true

nengo_loihi/builder.py
hunse: Because the string also gets used for naming the synapse, which is then used to find the synapse by name when we're making connections.

nengo_loihi/tests/test_interneurons.py
hunse: I've actually changed the test to have different functions for each node now, so it's more discerning.

nengo_loihi/builder.py
```diff
@@ -31,6 +31,250 @@
...
+    The (private) script for configuring these values can be found at:
+    https://gl.appliedbrainresearch.com/abr/nengo-loihi-sandbox
+    """
+
+    def __init__(self, **kwargs):
+        super(Preset10Interneurons, self).__init__(10, **kwargs)
+
+    def fix_parameters(self):
+        # Parameters determined by hyperopt
+        assert self.n == 10
```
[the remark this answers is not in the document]
hunse: I got rid of `fix_parameters`, since it seemed easier to just do things in the constructor. However, this is the one change that I can't do now. I don't think it's huge, but it does mean there's some overlap between the two constructors.

hunse:
Ok, I've made almost all the requested changes (except one, where removing a function precluded doing it).

Two things left to do are make the name change to DecoderNeurons, and test this on the chip (some of the tolerances on the new tests might have to be adjusted).

nengo_loihi/builder.py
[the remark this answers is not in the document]
tbekolay: Maybe `n_pairs` or `pairs_per_dimension`? I'm not okay with the one-character `n`.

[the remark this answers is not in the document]
tbekolay: I'd also be okay with `size` I think

tbekolay:
I pushed a fixup (8acfa6549d81045bc35) to change the class hierarchy around a little bit (simplified to remove `EqualDecoderDecodeNeurons`, as it seemed to only be there in order to have the `Preset` classes not have a `get_ensemble` method), can you take a look and let me know if I missed / misinterpreted anything @hunse? Tests are all still passing on emulator. I'll add tests for not currently covered lines later today.

hunse:
Ok, it was actually pretty easy to have them use the same `get_cx` function. I haven't gotten the `get_ensemble` to work yet; I'm going to look for a few more minutes and see.

hunse:
I've fixed up `get_ensemble`, too. It still doesn't work for `pairs_per_dim > 1`, for the reason noted in the code, but if we resolve that issue then I think it should work.

I'm happy with this, now!

tbekolay:
> I'm happy with this, now!

Ditto, the classes look great. I'll get started testing the diff lines not yet covered.

tbekolay:
> I think I would prefer `%0.3g` or `%0.2e`

Sure, I'll change to `%0.3g`

tbekolay:
The rebase was not difficult, but did require some changes, and I wrote a changelog entry and a consolidated commit message. Will wait for @hunse to sign off on the changelog entry and commit message before merging.

hunse:
The changelog entry looks good. The only error in the commit message is that NoisyDecodeNeurons was actually the default behaviour before this commit for Ensemble->Ensemble connections. The OnOffDecodeNeurons are only used to generate input spikes, both before and after this commit.

hunse:
I added a docstring for `noise_exp` so that it's clear that it's a base-10 float.

hunse:
There's one test that I've found failing on the chip so far, which is `test_communication.test_neurons2node`. The idea of the test is that you've got a sine wave input to an ensemble, whose neurons then connect out to a node which records the neural data. It's basically to test that a neuron->node connection works fine.

What's actually asserted, though, is that the neurons with positive encoders only fire for the positive part of the sine wave, and neurons with negative encoders only fire for the negative part. It seems like in this branch, on the chip, there are two stray spikes that happen in positive-encoder neurons during the negative part of the sine wave. Otherwise, the behaviour is identical between emulator and chip, and exactly as expected.

Really, I don't think this is a big cause for concern, and we could just have slightly looser tolerances on the test, since everything is behaving pretty-much as it should. But, I am annoyed that the emulator is not matching the chip perfectly here, and I'd like to look a bit more into that.

hunse:
Ok, so what seems to be happening here is that when we do spike probes with `precompute=False` (i.e. in snips), we probe the voltage and look for voltages that equal the refractory period * 128 (since Loihi uses the voltage to count the refractory period after a spike). In this test, with the new interneurons, it seems like two voltage values have just happened to hit this magic number (384 in this case) and we interpret that as a spike in our probe. Thus, it looks like the neuron has spiked, even though it hasn't.

If I run the test with `precompute=True` then it passes, because the Loihi spike probes do not have this problem.

So this is definitely unrelated to this PR, and I don't think we should try to fix it here. I would suggest adjusting the tolerances of the test to allow these couple of stray spikes, and make an issue for this.

hunse:
I made issue #170 to track this problem, and changed the test to use `precompute=True` for now. Hopefully everything runs on the chip now!

hunse:
I ran all the `nengo_loihi` tests on the chip, and they all pass for me.

I also ran the `nengo` tests on the chip, and they also appear to pass. I just want to double-check that it's only 70 tests that we expect to pass. (I got `70 passed, 508 skipped, 14 deselected, 198 xfailed` on our server, and `70 passed, 645 skipped, 14 deselected, 198 xfailed` on the Intel server, with no failures on either. I'm not sure why there's the discrepancy in numbers skipped; they're both using the same Nengo version, though one does look like it's from pip and the other from git.)

tbekolay:
> I'm not sure why there's the discrepancy in numbers skipped; they're both using the same Nengo version, though one does look like it's from pip and the other from git.

I think if you install from pip, you don't get the examples, so that might explain why it has fewer tests. 70 passing tests sounds about right.

All else sounds good, I'll do a quick double check of this then merge.

PR #133, opened 2018-10-30
Convnet prep

hunse:
I realized that there's a number of fixes as part of the convnet stuff that are useful by themselves, and would be helpful for the work I'm doing now. Since it still could be a little while before the actual convnet PR is merged, I figured that this could be a separate PR. A lot of it is not directly related to convnets, it's just bugs I found or things I needed to fix as part of doing that work.

### DIFF
--- a/.ci/test.sh
+++ b/.ci/test.sh
@@ -14,7 +14,7 @@ function usage {
 
 if [[ "$COMMAND" == "install" ]]; then
     conda install --quiet matplotlib mkl numpy scipy
-    pip install coverage pytest
+    pip install coverage 'pytest<4'
     pip install -e .
 elif [[ "$COMMAND" == "run" ]]; then
     coverage run -m pytest nengo_loihi -v --duration 20 --plots && coverage report
--- a/nengo_loihi/loihi_api.py
+++ b/nengo_loihi/loihi_api.py
@@ -305,7 +305,7 @@ class Core(object):
         a0 = 0
         if len(self.synapse_axons) > 0:
             last = next(reversed(self.synapse_axons))
-            a0 = self.synapse_axons[last][-1]
+            a0 = self.synapse_axons[last][-1] + 1
         idx_mult = 2 if synapses.tracing else 1
         idxs = [a0 + idx_mult*i for i in range(synapses.n_axons)]
         self.synapse_axons[synapses] = idxs
--- a/nengo_loihi/loihi_api.py
+++ b/nengo_loihi/loihi_api.py
@@ -290,10 +290,6 @@ class Core(object):
         self.vthProfiles.append(vth_profile)
         return len(self.vthProfiles) - 1  # index
 
-    def add_synapse_fmt(self, synapse_fmt):
-        self.synapseFmts.append(synapse_fmt)
-        return len(self.synapseFmts) - 1  # index
-
     def add_stdp_pre_cfg(self, stdp_pre_cfg):
         self.stdpPreCfgs.append(stdp_pre_cfg)
         return len(self.stdpPreCfgs) - 1  # index
@@ -303,8 +299,7 @@ class Core(object):
                    for synapses in group.synapses)
 
     def add_synapses(self, synapses):
-        synapse_fmt = synapses.synapse_fmt
-        synapse_fmt_idx = self.add_synapse_fmt(synapse_fmt)
+        synapse_fmt_idx = self.get_synapse_fmt_idx(synapses.synapse_fmt)
         self.synapse_fmt_idxs[synapses] = synapse_fmt_idx
 
         a0 = 0
@@ -328,6 +323,13 @@ class Core(object):
 
     def get_synapse_fmt(self, synapses):
         return self.synapseFmts[self.synapse_fmt_idxs[synapses]]
+
+    def get_synapse_fmt_idx(self, synapse_fmt):
+        try:
+            return self.synapseFmts.index(synapse_fmt)
+        except ValueError:
+            self.synapseFmts.append(synapse_fmt)
+            return len(self.synapseFmts) - 1  # index
 
 
 class Profile(object):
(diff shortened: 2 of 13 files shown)


hunse:
I was going to ask @tcstewar for a review, but I realized that @drasmuss didn't actually work on any of these commits, so he could review too. So I assigned you both, and you can decide who wants to do it :)

tbekolay:
Since it seems like having this in before #121 makes sense, I'm looking at this now. Going to fixup the fixups commits first because the full diff history is not useful, have to go commit by commit.

nengo_loihi/simulator.py
```diff
@@ -142,7 +142,7 @@ class Simulator(object):
     unsupported = []
 
     def __init__(self, network, dt=0.001, seed=None, model=None,  # noqa: C901
-                 precompute=False, target=None):
+                 precompute=False, target=None, loihi_args=None):
```
tbekolay: I really would like to limit the amount of args you can pass to the simulator. Everyone using Nengo Loihi is going to interact with the `Simulator`, so its interface needs to be as polished as possible. I feel like only devs are going to really want to futz with args passed to `LoihiSimulator`, and I believe the max spikes per step thing is removed in some PR in the queue, so figuring out an external interface to it isn't necessary.

Could we drop this commit from the PR? If you feel strongly about keeping it, then maybe have it be a separate PR?

nengo_loihi/loihi_cx.py
```diff
@@ -421,15 +421,24 @@ def format(self, **kwargs):
         self.synapse_fmt.set(**kwargs)
 
 
+class Spike(object):
+    __slots__ = ['axon_id', 'atom']
+
+    def __init__(self, axon_id, atom=0):
```
tbekolay: Could you add a docstring to this? Mostly I'm curious what an `atom` is.

nengo_loihi/loihi_cx.py
```diff
@@ -444,6 +453,16 @@ def axon_slots(self):
         """The total number of axonCfg slots used by all axons."""
         return self.slots_per_axon * self.n_axons
 
+    def map_cx_spikes(self, cx_idxs):
+        axon_idxs = (self.cx_to_axon_map[cx_idxs]
```
tbekolay: The logic for these three assignments is hard to read IMO ... is it ever the case that one of the three member variables will be defined, but not the other two? If it's all or none, then I think one if statement would be more clear than three if statements (one per variable).

nengo_loihi/loihi_cx.py
```diff
@@ -356,11 +356,11 @@ def size(self):
         return sum(len(w) for w in self.weights)
 
     def bits(self):
-        return sum(self.synapse_fmt.bits_per_axon(len(w))
+        return sum(self.synapse_fmt.bits_per_axon(w.size)
```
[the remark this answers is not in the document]
hunse: Generally, `CxSynapses.weights` is a list of ndarrays, but in practice in the past those ndarrays have always been 1-D, so this code has worked. However, with population synapses we want them to be 2-D, so this gets ready for that.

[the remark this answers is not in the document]
hunse: So it wasn't wrong, it just wasn't written as generally as I would've liked. If you want to be nit-picky, you could argue that it was fine before and this change should actually go in the `conv2d-master` PR.

nengo_loihi/loihi_api.py
```diff
@@ -304,7 +304,10 @@ def n_synapses(self):
 
     def add_synapses(self, synapses):
         synapse_fmt = synapses.synapse_fmt
-        synapse_fmt_idx = self.add_synapse_fmt(synapse_fmt)
+        try:
```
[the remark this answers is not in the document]
hunse: Hmm, that's fair. I'd want a different name for `add_synapse_fmt` then, because it wouldn't necessarily be adding it. Thoughts on a name?

nengo_loihi/loihi_api.py
```diff
@@ -313,7 +313,7 @@ def add_synapses(self, synapses):
         a0 = 0
         if len(self.synapse_axons) > 0:
             last = next(reversed(self.synapse_axons))
-            a0 = self.synapse_axons[last][-1]
+            a0 = self.synapse_axons[last][-1] + 1
```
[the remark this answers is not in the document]
hunse: Two ways: if `self.synapse_axons[last][-1]` is the last index we used, then we need to add 1 to get the next index. I came across it because I was looking at the compartments actually created and their indexes, and noticing that for groups with multiple synapses objects things weren't lining up.

We could write a hyper-specific test that looks at this. However, I consider this as one of a family of bugs where either: we overwrite things on the N2Board that we've already configured, or we miss configuring objects that we expect to have configured. Unfortunately, NxSDK does not make it easy to detect either of these things. I think the biggest start would be adding some "static" (not running anything on the chip) tests for `loihi_api` and `allocator`, which we've talked about doing, but has never been done.

nengo_loihi/loihi_cx.py
hunse: Somewhere down the line, I get rid of `axon_to_synapse_map` since we don't actually need it (essentially, `loihi_interface`/NxSDK figure this out themselves). `cx_to_axon_map` and `cx_atoms` are fairly independent. There are many situations where you might define `cx_to_axon_map`, to map different compartments to different axons, but they're not population axons so you don't use `cx_atoms`. Defining `cx_atoms` by itself would be much less common, because typically you're using population axons so that you can have multiple `cx`s running through the same axons, but in theory it's possible.

nengo_loihi/simulator.py
hunse: I need this for getting all larger convnets working, since they typically have high input spike bandwidth so I need to adjust `snip_max_spikes_per_step`. How about we add it for now, and then if/when we do remove `snip_max_spikes_per_step` then we can get rid of this? I can make it another PR if you'd prefer, but I don't think that's going to speed things along too much, because we'd still be waiting on that for the convnet stuff. I guess it would let us move on to the communication refactoring and learning.

nengo_loihi/loihi_cx.py
hunse: Yes, I'll do that.

.ci/test.sh
```diff
@@ -14,7 +14,7 @@ function usage {
 
 if [[ "$COMMAND" == "install" ]]; then
     conda install --quiet matplotlib mkl numpy scipy
-    pip install coverage pytest<4
```
hunse: Thanks. I was just encountering this in another branch.

Strangely, I upgraded my pytest to 4 on my machine, but I can't seem to reproduce it. So it's something a bit more particular about TravisCI.

nengo_loihi/loihi_interface.py
```diff
@@ -382,13 +382,13 @@ class LoihiSimulator(object):
...
-    snip_max_spikes_per_step : int, optional (Default: 50)
-        Maximum number of spikes that can be sent through
-        the nengo_io_h2c channel on one timestep.
     """
 
-    def __init__(self, cx_model,
-                 use_snips=True, seed=None, snip_max_spikes_per_step=50):
+    # Maximum number of spikes that can be sent through
+    # the nengo_io_h2c channel on one timestep.
+    snip_max_spikes_per_step = 50
```
hunse: This reduces functionality, IMO. We can no longer have multiple simulators with different values for this. Not that that is something that we'd ordinarily like to do, but definitely having it as a class attribute does not fit with my mental model as well.

I think the main reason that I'm hesitant is that I've had bad experiences having `CxSimulator.strict` being a class attribute. It means that in tests, we have to do a funky try/except to make sure that the value changes back after the test. Personally, I'm just not a fan of it at all.

I know we don't want to have a ton of `Simulator` arguments, but IMO there should be some way to get arguments down to the `LoihiSimulator` or `CxSimulator`.

tbekolay:
After some discussion, we settled on keeping it close to what it was before with additional documentation, which I added in 46a2db4ee43e60ab0. Let me know if that looks good @hunse

hunse:
LGTM!

PR #138, opened 2018-11-05
Move nahuku-connections.png to imgur

tbekolay:
Docs are included in the source distribution, and the image bulks up the distribution by 3.3 MB.

Since I was building the docs to test this anyhow, I also fixed a small issue in the nonlinear oscillator notebook.


hunse:
At the meeting we decided that we will purge the file. It will happen over the weekend to minimally effect people's work.

PR #140, opened 2018-11-12
nengo_dl builders for Loihi neurons

hunse:
This adds nengo_dl TensorFlow builders for the LoihiLIF and LoihiSpikingRectifiedLinear neuron types. This allows nengo_dl networks to train with tuning curves like those of the chip neurons, allowing for more accurate conversion of ANNs for the chip.

Based off #133

### DIFF
--- a/nengo_loihi/builder/nengo_dl.py
+++ b/nengo_loihi/builder/nengo_dl.py
@@ -1,0 +1,367 @@
+import logging
+import warnings
+
+import numpy as np
+
+from nengo_loihi.compat import HAS_DL
+from nengo_loihi.neurons import (
+    AlphaRCNoise,
+    discretize_tau_rc,
+    discretize_tau_ref,
+    LoihiLIF,
+    LoihiSpikingRectifiedLinear,
+    LowpassRCNoise,
+)
+
+if HAS_DL:
+    import nengo_dl
+    import nengo_dl.neuron_builders
+    import tensorflow as tf
+else:
+    # Empty classes so that we can define the subclasses even though
+    # we will never use them, as they are only used in the `install`
+    # function that can only run if nengo_dl is importable.
+    class nengo_dl:
+        class neuron_builders:
+            class LIFBuilder:
+                pass
+
+            class SpikingRectifiedLinearBuilder:
+                pass
+
+logger = logging.getLogger(__name__)
+
+
+class NoiseBuilder:
+    """Build noise classes in ``nengo_dl``.
+
+    Attributes
+    ----------
+    noise_models : list of NeuronOutputNoise
+        The noise models used for each op/signal.
+    """
+
+    builders = {}
+
+    def __init__(self, ops, signals, config, noise_models):
+        self.noise_models = noise_models
+        self.dtype = signals.dtype
+        self.np_dtype = self.dtype.as_numpy_dtype()
+
+    @classmethod
+    def build(cls, ops, signals, config):
+        """Create a NoiseBuilder for the provided ops."""
+
+        noise_models = [getattr(op.neurons, 'nengo_dl_noise', None)
+                        for op in ops]
+        model_type = (type(noise_models[0]) if len(noise_models) > 0
+                      else None)
+        equal_types = all(type(m) is model_type for m in noise_models)
+
+        if not equal_types:
+            raise NotImplementedError(
+                "Multiple noise models for the same neuron type is "
+                "not supported")
+
+        return cls.builders[model_type](ops, signals, config, noise_models)
+
+    @classmethod
+    def register(cls, noise_builder):
+        """A decorator for adding a class to the list of builders.
+
+        Raises a warning if a builder already exists for the class.
+
+        Parameters
+        ----------
+        noise_builder : NoiseBuilder
+            The NoiseBuilder subclass that the decorated class builds.
+        """
+        def register_builder(build_cls):
+            if noise_builder in cls.builders:
+                warnings.warn("Type '%s' already has a builder. Overwriting."
+                              % noise_builder)
+            cls.builders[noise_builder] = build_cls
+            return build_cls
+        return register_builder
+
+    def generate(self, period, tau_rc=None):
+        """Generate TensorFlow code to implement these noise models.
+
+        Parameters
+        ----------
+        period : tf.Tensor
+            The inter-spike periods of the neurons to add noise to.
+        tau_rc : tf.Tensor
+            The membrane time constant of the neurons (used by some noise
+            models).
+        """
+        raise NotImplementedError("Subclass must implement")
+
+
+@NoiseBuilder.register(type(None))
+class NoNoiseBuilder(NoiseBuilder):
+    """nengo_dl builder for if there is no noise model."""
+
+    def generate(self, period, tau_rc=None):
+        return tf.reciprocal(period)
+
+
+@NoiseBuilder.register(LowpassRCNoise)
+class LowpassRCNoiseBuilder(NoiseBuilder):
+    """nengo_dl builder for the LowpassRCNoise model."""
+
+    def __init__(self, ops, signals, *args, **kwargs):
+        super(LowpassRCNoiseBuilder, self).__init__(
+            ops, signals, *args, **kwargs)
+
+        # tau_s is the time constant of the synaptic filter
+        tau_s = np.concatenate([
+            model.tau_s * np.ones((op.J.shape[0], 1),
+                                  dtype=self.np_dtype)
+            for model, op in zip(self.noise_models, ops)])
+        self.tau_s = signals.constant(tau_s, dtype=self.dtype)
+
+    def generate(self, period, tau_rc=None):
+        d = tau_rc - self.tau_s
+        u01 = tf.random_uniform(tf.shape(period))
+        t = u01 * period
+        q_rc = tf.exp(-t / tau_rc)
+        q_s = tf.exp(-t / self.tau_s)
+        r_rc1 = -tf.expm1(-period / tau_rc)  # 1 - exp(-period/tau_rc)
+        r_s1 = -tf.expm1(-period / self.tau_s)  # 1 - exp(-period/tau_s)
+        return (1./d) * (q_rc/r_rc1 - q_s/r_s1)
+
+
+@NoiseBuilder.register(AlphaRCNoise)
+class AlphaRCNoiseBuilder(NoiseBuilder):
+    """nengo_dl builder for the AlphaRCNoise model."""
+
+    def __init__(self, ops, signals, *args, **kwargs):
+        super(AlphaRCNoiseBuilder, self).__init__(
+            ops, signals, *args, **kwargs)
+
+        # tau_s is the time constant of the synaptic filter
+        tau_s = np.concatenate([
+            model.tau_s * np.ones((op.J.shape[0], 1),
+                                  dtype=self.np_dtype)
+            for model, op in zip(self.noise_models, ops)])
+        self.tau_s = signals.constant(tau_s, dtype=self.dtype)
+
+    def generate(self, period, tau_rc=None):
+        d = tau_rc - self.tau_s
+        u01 = tf.random_uniform(tf.shape(period))
+        t = u01 * period
+        q_rc = tf.exp(-t / tau_rc)
+        q_s = tf.exp(-t / self.tau_s)
+        r_rc1 = -tf.expm1(-period / tau_rc)  # 1 - exp(-period/tau_rc)
+        r_s1 = -tf.expm1(-period / self.tau_s)  # 1 - exp(-period/tau_s)
+
+        pt = tf.where(period < 100*self.tau_s, (period - t)*(1 - r_s1),
+                      tf.zeros_like(period))
+        qt = tf.where(t < 100*self.tau_s, q_s*(t + pt), tf.zeros_like(t))
+        rt = qt / (self.tau_s * d * r_s1**2)
+        rn = tau_rc*(q_rc/(d*d*r_rc1) - q_s/(d*d*r_s1)) - rt
+        return rn
+
+
+class LoihiLIFBuilder(nengo_dl.neuron_builders.LIFBuilder):
+    """nengo_dl builder for the LoihiLIF neuron type.
+
+    Attributes
+    ----------
+    spike_noise : NoiseBuilder
+        Generator for any output noise associated with these neurons.
+    """
+    def __init__(self, ops, signals, config):
+        super(LoihiLIFBuilder, self).__init__(ops, signals, config)
+        self.spike_noise = NoiseBuilder.build(ops, signals, config)
+
+    def _rate_step(self, J, dt):
+        tau_ref = discretize_tau_ref(self.tau_ref, dt)
+        tau_rc = discretize_tau_rc(self.tau_rc, dt)
+        # Since LoihiLIF takes `ceil(period/dt)` the firing rate is
+        # always below the LIF rate. Using `tau_ref1` in LIF curve makes
+        # it the average of the LoihiLIF curve (rather than upper bound).
+        tau_ref1 = tau_ref + 0.5*dt
+
+        J -= self.one
+
+        # --- compute Loihi rates (for forward pass)
+        period = tau_ref + tau_rc*tf.log1p(tf.reciprocal(
+            tf.maximum(J, self.epsilon)))
+        period = dt * tf.ceil(period / dt)
+        loihi_rates = self.spike_noise.generate(period, tau_rc=tau_rc)
+        loihi_rates = tf.where(J > self.zero,
+                               self.amplitude * loihi_rates,
+                               self.zeros)
+
+        # --- compute LIF rates (for backward pass)
+        if self.config.lif_smoothing:
+            js = J / self.sigma
+            j_valid = js > -20
+            js_safe = tf.where(j_valid, js, self.zeros)
+
+            # softplus(js) = log(1 + e^js)
+            z = tf.nn.softplus(js_safe) * self.sigma
+
+            # as z->0
+            #   z = s*log(1 + e^js) = s*e^js
+            #   log(1 + 1/z) = log(1/z) = -log(s*e^js) = -js - log(s)
+            q = tf.where(j_valid,
+                         tf.log1p(tf.reciprocal(z)),
+                         -js - tf.log(self.sigma))
+
+            rates = self.amplitude / (tau_ref1 + tau_rc*q)
+        else:
+            rates = self.amplitude / (
+                tau_ref1 + tau_rc*tf.log1p(tf.reciprocal(
+                    tf.maximum(J, self.epsilon))))
+            rates = tf.where(J > self.zero, rates, self.zeros)
+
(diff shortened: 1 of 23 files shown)


nengo_loihi/neurons.py
```diff
@@ -220,3 +228,98 @@ def nengo_build_nif_rate(model, nif_rate, neurons):
...
+            J -= self.one
+
+            # --- compute Loihi rates (for forward pass)
+            period = tau_ref + self.tau_rc*tf.log1p(tf.reciprocal(
+                tf.maximum(J, self.epsilon)))
+            loihi_rates = (self.amplitude / dt) / tf.ceil(period / dt)
+            loihi_rates = tf.where(J > self.zero, loihi_rates, self.zeros)
+
+            # --- compute LIF rates (for backward pass)
+            if self.config.lif_smoothing:
```
drasmuss: Just brainstorming for the future, but it might be interesting to try a version of `loihi_rates` where the steps are smoothed a bit, rather than throwing out the steps altogether.  Same general idea as SoftLIF, but applying smoothing to all the discontinuities in the LoihiLIF function instead of just the threshold.

nengo_loihi/tests/test_neurons.py
```diff
@@ -83,3 +83,114 @@ def test_loihi_neurons(neuron_type, Simulator, plt, allclose):
...
     assert allclose(nengo_rates, ref, atol=atol, rtol=0, xtol=0)
     assert allclose(loihi_rates, ref, atol=atol, rtol=0, xtol=0)
+
+
+@pytest.mark.parametrize(
+    'neuron_type', [LoihiLIF(amplitude=1.0, tau_rc=0.02, tau_ref=0.002),
+                    LoihiLIF(amplitude=0.063, tau_rc=0.05, tau_ref=0.001)])
+def test_nengo_dl_loihi_lif(neuron_type, plt, allclose):
+    pytest.importorskip('nengo_dl')
+    import nengo_dl
```
drasmuss: `importorskip` actually returns the imported module, so you can do `nengo_dl = pytest.importorskip("nengo_dl")`

nengo_loihi/tests/test_neurons.py
```diff
@@ -218,3 +218,60 @@ def test_nengo_dl_neurons(neuron_type, plt, allclose):
...
+    with nengo_dl.Simulator(model, dt=dt) as sim:
+        input_data = {u: np.tile(x[None, None, :], (500, 1, 1))}
+        outputs = {ap: None}
+        y = sim.run_batch(input_data, outputs, training=True)[ap]
+        y = y[:, 0, 0, :]
+
+    ymean = y.mean(axis=0)
+    y25 = np.percentile(y, 25, axis=0)
+    y75 = np.percentile(y, 75, axis=0)
+
```
drasmuss: Can you add some assertions to this test?

nengo_loihi/neurons.py
```diff
@@ -231,7 +235,65 @@ def nengo_build_nif(model, nif, neurons):
 
 
 if nengo_dl is not None:  # noqa: C901
+    class TFSpikeNoiseGenerator(object):
```
drasmuss: Is the reason that we can't use Nengo's ensemble noise parameter for this that the noise is dependent on the input current?  Assuming that I'm understanding that correctly, maybe we should consider reworking the ensemble noise builder so that we could support that.

nengo_loihi/neurons.py
hunse: That's an interesting idea. The main reason for the SoftLIF is to smooth the derivative, and here I'm not using the steps at all in the derivative, so the motivation would be a bit different. In practice, though, I think the steps of the tuning curve would get smoothed somewhat in practice due to noise in the network. If you train with noise (which I've been doing for some of the keyword network stuff), then you'll actually get some smoothing of the steps like you're describing, however it will be probabilistic. Doing it deterministically like you suggest would have the benefit of making it more reliable for the network, and therefore allow it to take advantage of it better. You could even (theoretically) use a derivative that takes into account the smoothed steps (maybe that's what you were thinking all along).

The downsides are that you'd have another hyperparameter to tune (you could try to find a good value by estimating the amount of noise on the signals into each layer, and choosing the amount of smoothing to mimic this). The other problem is that I think a simple analytic equation for it (which I think is what we need for TensorFlow) might be difficult to find, keeping in mind that we need it to be relatively easily computable on a GPU. The SoftLIF wasn't trivial to come up with, and that is a considerably simpler problem with only one discontinuity. Applying the same type of SoftLIF smoothing to each of the discontinuities in the stepped Loihi tuning curve would result in an overly-complex function, I think. Maybe there's a way to do it efficiently, though.

nengo_loihi/tests/test_neurons.py
hunse: I don't like testing things with noise. I can never think of good assertions.

Also, I think I'm going to take the noise stuff out for now. I need to look more at efficiency, since I've learned that the `tf.switch` statement is not lazy (it evaluates both branches), though maybe it is lazy if the condition is a constant. Do you know anything about this @drasmuss?

nengo_loihi/neurons.py
hunse: The ensemble noise is noise on the input current/voltage. This is the opposite kind of noise, noise on the output. It really only makes sense in the case of rate neurons, but I agree, it would be nice if Nengo supported it in some way (though I also don't want to clutter up the API, since I think this is only of interest to a small subset of users).

nengo_loihi/neurons.py
drasmuss: > You could even (theoretically) use a derivative that takes into account the smoothed steps (maybe that's what you were thinking all along).

Yeah that's what I was thinking.  The nice thing about this implementation is that we're using Loihi-like neurons on the forwards pass, but it would be nice if we could use them on the backwards pass as well.  You're right though that I'm not sure what a compact expression for that would be.

nengo_loihi/tests/test_neurons.py
drasmuss: I know it's not lazy if the condition is a placeholder.  It might be if the condition were a constant, but in that case we could just get rid of the conditional all together.

nengo_loihi/neurons.py
drasmuss: Since this is just for rate neurons, and we know the nonlinearity, it seems like there should be some equivalent noise on the input side that would produce the same effect as noise on the output.  I haven't thought about that too carefully though 🤷‍♂️ 

In terms of API, maybe that'd look like a `noise` parameter on the `Connection`?  Definitely not a high priority at the moment though, agreed.

nengo_loihi/tests/test_neurons.py
hunse: Basically, I'm using the conditional to select which neurons to update with a particular noise model; essentially it's an indexing operation. Take a look at it here: https://github.com/nengo/nengo-loihi/pull/140/files#diff-e5b03ba0ff51ec913bbc99cf1732ac8bR286. Do you know how to do it better?

nengo_loihi/neurons.py
hunse: So, for inputs greater than the firing threshold, the tuning curve is a one-to-one function. That means that it has a well defined inverse that you could apply to the output noise to get equivalent input noise. However, it's going to be messy, and you'll end up undoing all that computation when you pass things through the nonlinearity anyway, so why do it? As you might have noticed, the noise equations I'm using are quite specialized distributions (developed in my thesis), not a simple normal or uniform distribution.

You could put it on the connection, or you could allow Ensembles to have both input and output noise. I think the main use case is training in nengo_dl, though, so we might even want to consider some sort of config option that's nengo_dl specific.

nengo_loihi/tests/test_neurons.py
drasmuss: Or if you want to reduce the number of values for which you compute `alpha_rc_noise`, you could do something like
```
y = period + tf.scatter_nd(self.alpha_rc_noise.gather(kind_indices), kind_indices, ...)
```
where `kind_indices` are the neuron indices for which you want to compute noise.

nengo_loihi/neurons.py
drasmuss: > However, it's going to be messy, and you'll end up undoing all that computation when you pass things through the nonlinearity anyway, so why do it?

Well the advantage would be that you could use the existing Ensemble noise parameter, rather than having to hack things into the neuron model.  But agreed that it's not worth worrying too much about it for this one case.

hunse:
- [ ] Enforce same noise model for all grouped neuron operators

hunse:
A number of tests have starting failing for weird reasons. `test_learning.test_multiple_pes` fails sporadically; hopefully with #151 it will be more predictable.

There's also failing for the static and docs tests, because of weird problems with the setup for CI. Not sure if it's because something changed with packages in pip, or how the Travis servers are initialized, or what.

hunse:
Also, the codecov is bad because the new tests are `importorskip('nengo_dl')`, and it appears they're not getting run. But `nengo_dl` should be installed for the tests, I thought. I think it might have to do with the version (I might be using some newer `nengo_dl` features).

nengo_loihi/neurons.py
```diff
@@ -197,3 +269,354 @@ def nengo_build_nif_rate(model, nif_rate, neurons):
...
+
+        def generate(self, period, tau_rc=None):
+            assert tau_rc is not None
+            return self.tensorflow(period, self.tau_s, tau_rc)
+
+    class LowpassRCNoiseBuilder(RCNoiseBuilder):
+        """nengo_dl builder for the LowpassRCNoise model."""
+
+        @classmethod
+        def tensorflow(cls, period, tau_s, tau_rc):
```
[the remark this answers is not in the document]
hunse: There was a docstring under the `RCNoiseBuilder` class, which fa90c22 gets rid of. In fact, it looks like this function no longer exists at all.

The reason that I had it was that originally, I was hoping to have a noise class that could do many different noise models together, and this kind of general function might be useful. But that's not happening anymore, so I think I agree that it makes sense to get rid of it.

nengo_loihi/neurons.py
```diff
@@ -197,3 +269,354 @@ def nengo_build_nif_rate(model, nif_rate, neurons):
 @NengoBuilder.register(NIF)
 def nengo_build_nif(model, nif, neurons):
     return build_lif(model, nif, neurons)
+
+
+class NeuronOutputNoise(object):
```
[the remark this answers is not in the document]
hunse: It was mostly just to group these classes that describe models for noise on neuron outputs. That way, in docstrings, I had something to say as a type parameter. And I could have the general docstring letting people know what these classes do, so I wouldn't have to specify that general stuff for each class (of which there are currently only two). But that's the only role I have for it right now. It's more a conceptual thing than a functional one.

nengo_loihi/neurons.py
```diff
@@ -321,35 +321,41 @@ class NoiseBuilder(object):
...
-        def __init__(self, ops, signals, config, models):
-            self.models = models
+        def __init__(self, ops, signals, config, noise_models):
+            self.noise_models = noise_models
             self.dtype = signals.dtype
             self.np_dtype = self.dtype.as_numpy_dtype()
-            self.np_ones = [
-                np.ones((op.J.shape[0], 1), dtype=self.np_dtype) for op in ops]
+
+            if hasattr(noise_models[0], "tau_s"):
```
hunse: I don't like this. I always bristle at `hasattr`/`getattr`, even though they do have their uses sometimes. Avoiding this is why I created the RCNoiseModel class initially. But maybe it's not worth having another class just to avoid it.

nengo_loihi/neurons.py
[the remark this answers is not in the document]
drasmuss: The docstring there was just "Generate TensorFlow code for this model type given parameters.", I was thinking more specifically about understanding the noise models here (e.g., what distribution is this function implementing, and why that distribution?).  For example, right now I can see all the math that is being implemented, but I can't see the reasoning underlying that math.  Basically a (very brief) introduction to why and how you developed these noise models, so that another developer coming along in the future could conceivably understand the "why" of the code without having to talk to you.  Maybe include a reference to the relevant section of your thesis?

nengo_loihi/neurons.py
drasmuss: Another option would be to duplicate this in the constructor of the two subclasses.  Not ideal to duplicate code like that, but it is basically a one-liner.  I don't have a strong preference either way.

nengo_loihi/neurons.py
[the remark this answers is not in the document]
drasmuss: Yeah I think I'm fine keeping it actually, I have found it useful having a unified reference object for documentation.

nengo_loihi/neurons.py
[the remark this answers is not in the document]
hunse: So should that docstring go under the noise model, or the noise model builder? Or maybe a bit of both? I definitely think that the noise model should have a bit more description and thesis reference. If I do that, do we still want something for the noise model builders?

nengo_loihi/neurons.py
hunse: I would duplicate it, I think. Where it is right now, it's not where I would expect to find that code if I knew I had to change it.

nengo_loihi/tests/test_neurons.py
```diff
@@ -118,3 +169,231 @@ def test_lif_min_voltage(Simulator, plt, allclose):
...
+    neurons.nengo_dl = neurons.tf = None
+
+    with nengo.Network() as net:
+        a = nengo.Ensemble(10, 1, neuron_type=LoihiLIF())
+        nengo.Probe(a)
+
+    with Simulator(net) as sim:
+        sim.step()
+
+    neurons.nengo_dl, neurons.tf = saved
```
hunse: Should we be putting this in a try/finally block or using mock, just to make sure if there's an error that these get reset and don't affect other tests?

hunse:
All LGTM now!

nengo_loihi/tests/test_neurons.py
```diff
@@ -118,3 +167,151 @@ def test_lif_min_voltage(Simulator, plt, allclose):
...
+
+    np.fill_diagonal(analytic, 0)
+    assert np.all(analytic == 0)
+
+    assert allclose(y, y_ref, atol=1e-3, rtol=1e-5)
+    assert allclose(dy, dy_ref, atol=1e-3, rtol=1e-5)
+    assert allclose(y_spikerate, y_ref, atol=1, rtol=1e-2)
+
+
+def test_no_nengo_dl(Simulator, monkeypatch):
```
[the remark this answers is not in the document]
drasmuss: It does affect the use of the `tf` or `nengo_dl` attributes within functions/classes inside `neurons.py`, which is what this test was designed to catch.  E.g., previously we had a bug where a function assumed that `tf` was installed, and it would crash if it wasn't, and this test caught that error.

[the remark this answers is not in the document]
tbekolay: Right, I only noticed after making this change that `tf` is used in the `discretize` functions which are used by some of the other classes. I'll add it back.

nengo_loihi/neurons.py
```diff
@@ -8,12 +8,11 @@
 import numpy as np
 
 try:
-    import nengo_dl
-    import nengo_dl.neuron_builders
     import tensorflow as tf
-except ImportError:  # pragma: no cover
-    nengo_dl = None
+    HAS_TF = True
+except ImportError:
```
hunse: This does not appear to be covered. I think we want to keep the `pragma: no cover`.

nengo_loihi/builder/nengo_dl.py
```diff
@@ -0,0 +1,369 @@
...
+        Raises a warning if a builder already exists for the class.
+
+        Parameters
+        ----------
+        noise_builder : NoiseBuilder
+            The NoiseBuilder subclass that the decorated class builds.
+        """
+        def register_builder(build_cls):
+            if noise_builder in cls.builders:
+                warnings.warn("Type '%s' already has a builder. Overwriting."
```
hunse: This warning is uncovered.

nengo_loihi/neurons.py
tbekolay: Removing the `pragma: no cover` was accidental, will add it back

nengo_loihi/tests/test_neurons.py
```diff
@@ -245,7 +251,7 @@ def test_nengo_dl_neuron_grads(neuron_type, plt, allclose):
     else:
         # use the derivative of y_med (the smoothed Loihi tuning curve)
         dy_ref = np.zeros_like(j)
-        dy_ref[j > 0] = neuron_type.amplitude / (j[j > 0]*tau_ref1 + 1)**2
+        dy_ref[j > 0] = neuron_type.amplitude / (j[j > 0]*tau_ref1 + 1) ** 2
```
hunse: PEP8 doesn't require space around all binary operators: https://www.python.org/dev/peps/pep-0008/#other-recommendations

I'm really not sure what PEP8 would recommend in this case. Because `**` is a "lower priority" operator here, you could argue that it should have these spaces. However, I think because I'm used to cases where it's higher priority (PEP8 clearly prefers `x**2 + 1` to `x ** 2 + 1`), I always do not put spaces around `**`. It looks really weird to me.

hunse: Maybe it's partly because `**` is already two characters, so there's a bit of "space" there. With the added spaces, it just seems like so much distance, especially because in math a superscript is small and so it feels even closer to whatever it's raising to a power.

hunse: I mean, I also don't agree with PEP8: I prefer `(a + b)*(c - d)` to `(a+b) * (c-d)`. So maybe it's correct to have the spaces here, but not in the `x**2 + 1` case. I guess I just like things to be fairly consistent, so I'm not always thinking "do I need spaces here or not"? Which is kind of the point of PEP8, but I feel like here it actually makes these decisions harder.

hunse:
Other than those two coverage points, the changes look good to me.

nengo_loihi/tests/test_neurons.py
tbekolay: Yeah, I prefer putting spaces around all operators so that I don't need to make any decisions about whether some equation will be easier to parse with certain spacing. Then since I do that, it seems more readable to me because I expect there to be spaces around all operators. I would be in favor of enforcing that with flake8, but for now the `**2` just seemed weird to me, but I can change it back, no biggie

drasmuss: Yeah I prefer spaces around all operators for the same reason.  It just removes an extra decision, and makes it easy to apply automatic formatting.

nengo_loihi/tests/test_learning.py
```diff
@@ -224,6 +224,47 @@ def test_multiple_pes(init_function, allclose, plt, seed, Simulator):
                         atol=0.05, rtol=0.05), "Target %d not close" % i
 
 
+def test_pes_deterministic(request, Simulator, seed, plt, allclose):
```
tbekolay: `plt` isn't used in this test and should be removed to avoid the empty plot that's generated.

PR #142, opened 2018-11-19
Update connections to work with new Transform API

drasmuss:
This gets all the tests passing with nengo master.  Note that this does not support the new Convolution transforms, it just brings the existing nengo-loihi functionality up-to-date with nengo core.  There are still several open PRs related to the nengo-loihi convolution implementation, so I figured I would wait for those to be resolved before trying to apply those changes.

Also note that right now I've done this in a non-backwards-compatible way (i.e. it does not work with nengo 2.8.0).  Since nengo-loihi is pretty "in development", requiring the most up-to-date version of Nengo (assuming there is a new nengo release before the next nengo-loihi release) seems reasonable.  But we could make it backwards compatible, it just adds a little extra complexity and maintenance burden.

TODO:
- [x] improve test coverage
- [x] update conv example notebook


nengo_loihi/splitter.py
```diff
@@ -263,13 +264,11 @@ def split_host_to_chip(networks, conn):
...
-        conn.transform,
-        n=conn.size_out,
-        d=conn.size_mid,
-        rng=np.random.RandomState(seed=seed))
+    weights = conn.transform.sample(rng=np.random.RandomState(seed=seed))
     if isinstance(conn.post_obj, nengo.Ensemble):
-        transform = transform / conn.post_obj.radius
+        weights = weights / conn.post_obj.radius
+    transform = copy.copy(conn.transform)
+    type(transform).init.data[transform] = weights
```
hunse: That looks weird. If there's no way to do that nicer, maybe we're missing some sort of functionality in `FrozenObject` or `Transform`.

nengo_loihi/builder.py
```diff
@@ -812,7 +814,10 @@ def signal_probe(model, key, probe):
     if kwargs is not None:
         if kwargs['function'] is not None:
             raise BuildError("Functions not supported for signal probe")
-        weights = kwargs['transform'].T / model.dt
+
+        assert isinstance(kwargs["transform"], Dense)
```
hunse: I would do a `BuildError` here rather than an assert, since a user could reasonably trip this.

nengo_loihi/builder.py
```diff
@@ -760,7 +761,8 @@ def conn_probe(model, probe):
             else:
                 input_dim = len(func[0])
         transform = kwargs['transform']
-        transform = np.asarray(transform, dtype=np.float64)
+        assert isinstance(transform, Dense)
```
hunse: I'd also do a BuildError here (I'm looking through this backwards :back:)

nengo_loihi/builder.py
```diff
@@ -517,8 +518,8 @@ def build_connection(model, conn):
     neuron_type = None
 
     # Sample transform if given a distribution
-    transform = get_samples(
-        conn.transform, conn.size_out, d=conn.size_mid, rng=rng)
+    assert isinstance(conn.transform, Dense)
```
hunse: Another place for a `BuildError`!

nengo_loihi/builder.py
drasmuss: I wasn't anticipating that we'd merge this in before supporting Convolutions, so these are just here as a temporary reminder that this assumes Dense transforms.

nengo_loihi/builder.py
drasmuss: Same as above.

nengo_loihi/builder.py
drasmuss: Same as above 😄

nengo_loihi/builder.py
hunse: Ok, sounds good!

drasmuss:
Rebased this to master, and added support for `nengo.Convolution` transforms.  All the tests are passing, but I'm not confident that that means things are working, as the test coverage is a bit sparse.  For example, it looks like a lot of the tests were set up as if they should test different conditions, but then those conditions aren't being tested (e.g. https://github.com/nengo/nengo-loihi/pull/142/commits/e39c54b34d3f9fb9fa749d6c9e69fee5ff4d72df#diff-2b5472730de7e72d3c4c167f986643d2R46, https://github.com/nengo/nengo-loihi/pull/142/commits/e39c54b34d3f9fb9fa749d6c9e69fee5ff4d72df#diff-2b5472730de7e72d3c4c167f986643d2R285, https://github.com/nengo/nengo-loihi/pull/142/commits/e39c54b34d3f9fb9fa749d6c9e69fee5ff4d72df#diff-2b5472730de7e72d3c4c167f986643d2R359, https://github.com/nengo/nengo-loihi/pull/142/commits/e39c54b34d3f9fb9fa749d6c9e69fee5ff4d72df#diff-2b5472730de7e72d3c4c167f986643d2R397).  But I'm not sure if those un-tested conditions are expected to be working or not.  So need @hunse to look it over, since he wrote the tests and has used the convolution code the most.

nengo_loihi/conv.py
```diff
@@ -156,7 +156,11 @@ def build_conv2d_connection(model, conn):
     assert conn.post_slice == slice(None)
 
     gain = model.params[conn.post_obj.ensemble].gain
-    assert np.all(gain == gain[0]), "All gains must be the same"
+    if not np.all(gain == gain[0]):
+        # TODO: support this
+        raise NotImplementedError(
```
hunse: I'm not sure how we could ever support this. If each neuron has a different gain, then there's no way to fold those gains into the convolution weights (since the essence of convolution is that neurons at different locations have the same weights).

nengo_loihi/conv.py
```diff
@@ -669,42 +227,59 @@ def conv2d_loihi_weights(conv2d_transform):
...
-                    + nf*np.arange(wmask_j.sum())[:, None]
-                    + np.arange(nf)
-                ).reshape(nc, -1)
+                        np.zeros((inp_shape.n_channels, 1, 1, 1), dtype=int)
+                        + (output_cols * transform.n_filters
+                           * np.arange(wmask_i.sum())[:, None, None])
+                        + transform.n_filters * np.arange(wmask_j.sum())[:,
+                                                None]
+                        + np.arange(transform.n_filters)
+                ).reshape(inp_shape.n_channels, -1)
```
hunse: Personally, I find this harder to read than how it was before. For me, I guess it feels like math; I find math much easier to read when it's all one-letter variable names (possibly with a short subscript, as math is typically written). This new code looks much more intimidating to me, and it's much less clear what's going on. The pattern where each line is being broadcasted to a different row of the matrix gets lost.

nengo_loihi/conv.py
drasmuss: I'll change it to a validation error instead, to indicate that it probably won't change anytime in the near future.

nengo_loihi/conv.py
drasmuss: For me it's the opposite, I find the single-letter variable code very hard to read.  E.g. if I look at `nyf` I have no idea what that is unless I trace its origin all the way back through the code, but `output_cols` gives me a pretty good idea what that variable represents.  We might need a tie-breaker vote on this one 😄 .

nengo_loihi/conv.py
drasmuss: Done!

hunse:
- [ ] The `xfail` tests in `test_conv.py` should all work on the emulator, and currently they're not. I think it's just syntax changes related to this PR that were never fixed. So a reminder to do that before we merge.

nengo_loihi/conv.py
hunse: I've made it clearer to me by making `inds0`, `inds1` etc. variables for each element in the sum, and then adding them together at the end. Take a look and let me know what you think.

nengo_loihi/splitter.py
hunse: It might be nice to have a `copy` function on `FrozenObject` that let's you make a copy, but pass any constructor arguments to change those values. Could be useful for users if they've got some object like a neuron type, and want to play around with some specific parameter while keeping everything else constant. I'll make a Nengo issue, and we can leave this as-is for now.

hunse: nengo/nengo#1509

nengo_loihi/builder/connection.py
```diff
@@ -104,10 +139,10 @@ def build_connection(model, conn):
     eval_points = None
     solver_info = None
     neuron_type = None
+    post_slice = conn.post_slice
 
-    # Sample transform if given a distribution
-    transform = get_samples(
-        conn.transform, conn.size_out, d=conn.size_mid, rng=rng)
+    # sample transform (if using a distribution)
+    transform = sample_transform(conn, rng=rng)
```
hunse: If we sample the transform here, then we can't sample it separately in `build_decoders` if we have a weight solver, since then we'll definitely get a different sampled transform if we use weight solvers as if we don't.

nengo_loihi/compat.py
```diff
@@ -0,0 +1,38 @@
...
+        return conn.transform.sample(rng=rng)
+
+else:
+    nengo_transforms = None
+    from nengo.dists import get_samples as _get_samples
+
+    nengo.solvers.Solver.compositional = True
+    nengo.solvers.LstsqDrop.compositional = False
+    nengo.solvers.LstsqL1.compositional = False
+    nengo.solvers.Nnls.compositional = False
```
hunse: One danger to this is it would affect other packages/places someone is using nengo. It's quite unlikely that that would actually be a problem, but I do always worry about monkey patching like this. It does make things much cleaner, though, so I like that aspect.

.travis.yml
```diff
@@ -7,13 +7,20 @@ notifications:
...
     - PYTHON="3.5.2"
+    - NENGO_VERSION="nengo"
 
 matrix:
   include:
-    - env: MODE="emulator"
-    - env: MODE="hardware"
+    - stage: basic
+      env: MODE="emulator"
     - env: MODE="static"
```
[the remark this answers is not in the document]
drasmuss: My thinking was that the "static" build is quick to run and generally quick to fix, so it makes sense to have it as part of the early filtering (i.e., if things aren't passing static checks we can determine that early and not use up other resources, and if someone does want to run the Advanced stage it should be pretty quick to fix the static errors and re-run).

nengo_loihi/compat.py
drasmuss: Yeah, in this case since this is only an internal builder attribute (so we know how and where it is being used), the risk seemed low enough to me.  The only case this would cause a problem is if a) someone is writing their own code that uses the compositional attribute (writing their own build functions or something?), and b) they don't want those attributes to have the values defined in the solver (for some reason).  And this person has to be using an old version of nengo, where the `compositional` attribute didn't even exist.

nengo_loihi/builder/connection.py
drasmuss: Hmm yeah let me see if there's a nice way to clean that up.

drasmuss: I just ended up explicitly passing the transform to build_decoders (which I know you had in at some point and then I took out 😜)

.travis.yml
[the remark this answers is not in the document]
hunse: FYI, it looks like we can configure gitlint for what words it errors on in titles: https://jorisroovers.github.io/gitlint/rules/. By default, it just doesn't like "WIP". So if you ever don't want something to run full tests, have a WIP commit. TMP and NOMERGE and things like that will not trigger gitlint.

nengo_loihi/builder/connection.py
hunse: Yeah, that's why I did it that way :)

PR #143, opened 2018-11-20
Documentation updates

drasmuss:
Switches the documentation to the new theme, and adds versioning to the build process.


tbekolay:
I think this is also something we should have in 0.4.0. I looked it over and I think this is basically good to go, except that we should add in the Nengo Loihi logo from https://github.com/nengo/design/pull/7 as currently there's no other way to get back to the index page. Aside from that, is there anything else you want to do in this branch @drasmuss ?

tbekolay:
Also a heads up that I rebased this!

drasmuss:
Pushed some fixups, and added a logo and favicon from nengo/design#7.  Took a look at the built docs and they look good to me!

docs/conf.py
```diff
@@ -74,18 +71,15 @@
...
-    "base_url": "https://www.nengo.ai/nengo-loihi",
-}
+html_theme = "nengo_sphinx_theme"
 
 html_title = "Nengo Loihi {0} docs".format(release)
 htmlhelp_basename = 'Nengo Loihi'
 html_last_updated_fmt = ''  # Suppress 'Last updated on:' timestamp
 html_show_sphinx = False
+html_favicon = os.path.join("_static", "favicon.ico")
+html_logo = os.path.join("_static", "logo.png")
```
tbekolay: Hmm, I was envisioning using `nengo_loihi/full-light.svg` rather than the square logo in png form ... so looking similar to https://www.nengo.ai/design 

I see that NengoDL uses the big square one, so I wonder if this is something we need to standardize or if we allow different projects to choose differently?

drasmuss: That's kind of a holdover from the previous NengoDL theme, where it already said "NengoDL" in the side-bar so having a logo say that as well seemed a bit redundant.  I think I still slightly prefer the simpler aesthetic of the single N logo, but it's not a strong preference.

drasmuss: Actually I'd probably downgrade my "slight preference" to "no preference".

tbekolay: Made a vote for this in https://github.com/nengo/design/issues/9.

drasmuss:
Added a fixup to change the logo and adjust the width (with the new nengo-sphinx-theme release).

drasmuss:
Looks good to me.  I added a commit to apply the patch from https://github.com/nengo/nengo-bones/pull/3/files.  Not strictly necessary since we don't allow PR builds in this repo, but doesn't hurt.

PR #151, opened 2018-12-07
Ensure things happen in consistent order

hunse:
- Splitter preserves object order from original network.
- Host/chip receivers are always called in same order.

This makes sure that seeds are always set the same as well. Right now, setting a seed on a network does not actually guarantee it will be the same from one run to the next, because the splitter dictionaries change the order that everything gets built in in a non-deterministic way.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -44,6 +44,11 @@ Release history
   like integrators.
   (`#124 <https://github.com/nengo/nengo-loihi/pull/124>`_,
   `#114 <https://github.com/nengo/nengo-loihi/issues/114>`_)
+- Ensure things in the build and execution happen in a consistent order from
+  one build/run to the next (by using ``OrderedDict``, which is deterministic,
+  instead of ``dict``, which is not). This makes debugging easier and seeding
+  consistent.
+  (`#151 <https://github.com/nengo/nengo-loihi/pull/151>`_)
 
 0.4.0 (December 6, 2018)
 ========================
--- a/nengo_loihi/inputs.py
+++ b/nengo_loihi/inputs.py
@@ -2,6 +2,7 @@ from __future__ import division
 
 from nengo import Node
 from nengo.exceptions import SimulationError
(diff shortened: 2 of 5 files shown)


nengo_loihi/loihi_cx.py
```diff
@@ -1164,12 +1166,12 @@ def receive(self, t, x):
 class ChipReceiveNode(nengo.Node):
     """For receiving host->chip messages"""
 
-    def __init__(self, dimensions, size_out):
+    def __init__(self, dimensions, size_out, **kwargs):
```
tbekolay: I think `**kwargs` is confusing here; for me, I don't really remember what other parameters there are to `Node`. If I look it up and see `size_in`, I'll get a `TypeError` if I try to specify it. I think we should bite the bullet and expose `label` and `seed` (or just `label` if seeds are irrelevant here).

nengo_loihi/loihi_cx.py
```diff
@@ -1190,6 +1192,7 @@ def collect_spikes(self):
 
 class ChipReceiveNeurons(ChipReceiveNode):
     """Passes spikes directly (no on-off neuron encoding)"""
-    def __init__(self, dimensions, neuron_type=None):
+    def __init__(self, dimensions, neuron_type=None, **kwargs):
```
tbekolay: Here also, explicitly include `label` and (perhaps) `seed`.

hunse:
I also can't think of a test that is non-trivial. Maybe #70 will test some of this, though also the point of #70 was to make sure that things have the same seeds as Nengo no matter what order they're built in.

I've addressed the comments and added a changelog.

tbekolay:
From the review: we should add a test that splits a model multiple times and ensures that their order in the split networks is the same.

hunse:
I will write this test.

hunse:
Ok, I've added a test. I've tested that it can distinguish between before this PR and after it.

PR #156, opened 2018-12-13
Neurons to ensemble connections with interneurons

hunse:
An alternative to #71.

In #71, I feel like I cheated, and specifically implemented neurons->ensemble connections that only have a scalar transform. In that case, the dimensionality of the post-ensemble has to be the same as the pre neurons.

In this PR, I do it in a more general way, by including interneurons in the case when we're connecting from neurons to ensemble. This allows us to put arbitrary transforms in there.

HOWEVER, I did find that for the case where we have neurons->ensemble connections of the first type, with scalar transforms, this new method does not work well. I think that's because each group of interneurons (i.e. each dimension in the interneurons population) is being driven only by one input neuron, so its input is very noisy and it just doesn't perform well. Furthermore, this would only work for small numbers of input neurons (less than 50 or so), since we use 20 interneurons to represent each dimension so for greater numbers of pre neurons we'd just run out of neurons on a single core for the interneurons.

Anyway, I feel like this is not a big limitation, since if you have the encoders for the post population, you can just do a neurons->neurons connection with the encoders as the transform to get behaviour (without interneurons) like I had implemented before. It can be a bit annoying to get the encoders, since essentially it means you have to generate them yourself at model creation time, but this isn't a huge burden.

Also, this is much simpler than #71 was, since there I had to allow for different encoders depending on whether we were connecting from interneurons (i.e. another ensemble) or neurons.

### DIFF
--- a/.nengobones.yml
+++ b/.nengobones.yml
@@ -282,8 +282,6 @@ setup_cfg:
         inaccurate
 
       # builder inconsistencies
-      test_connection.py:test_neurons_to_ensemble*:
-        transform shape not implemented
       test_connection.py:test_transform_probe:
         transform shape not implemented
       test_connection.py:test_list_indexing*:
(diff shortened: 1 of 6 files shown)


tbekolay:
Rebased this to current master.

Issue #155, opened 2018-12-10
Scaling n_neurons past ~200 does not improve accuracy


arvoelke:
There seems to be a critical noise floor (~5% error) that can't be overcome by adding more neurons to a single layer encode-decode. The floor is already hit around 200 neurons. This could have to do with synchrony in the spikes (i.e., a lack of heterogeneity, causing two neurons to be no different than one with twice the spike height)? This synchrony could be coming from the discretization of the neuron model in time, or the quantization of its variables and time-constants in space? Or systematic differences between `loihi_rates` and the actual spike rates?

I've tested this (on master) against a wide range of L2-regularization constants (from `1e-5` to `1e0`). With standard Nengo, you can always continue to improve performance by increasing `n_neurons` and dropping `reg` according to the spike-noise (`O(1/sqrt(n))`), but it doesn't seem to be the case here

Is there any way to break the ~5% floor?

![loihi-scaling-2](https://user-images.githubusercontent.com/5420057/49764405-558a6b80-fc9d-11e8-8aab-5be2932f7197.png)

Of relevance to #126: uncommenting the configuration to the `max_rates` and `intercepts` defaults below makes everything worse by raising the noise floor to ~17% (not shown). I believe this is because it increases the level of spike synchrony by narrowing the range of the response curves at `1`. And so further investigation might be required to determine the best defaults.

On the other hand, lowering `dt` does seem to make things _slightly_ better.

![loihi-dt](https://user-images.githubusercontent.com/5420057/49766391-27a82580-fca3-11e8-8321-1bd028c1d14e.png)

I've also tried partitioning up to `2000` neurons across many sub-ensembles, and solving the decoder optimization problem across all ensembles simultaneously using my `VirtualEnsemble` network, and that actually shows performance degrade past `1000` neurons (code ommitted, for brevity).

![virtual_ensemble](https://user-images.githubusercontent.com/5420057/49767447-77d4b700-fca6-11e8-849b-c70718152bd0.png)

```python
from collections import defaultdict

import numpy as np
import matplotlib.pyplot as plt
import seaborn as sns
from pandas import DataFrame

import nengo
from nengo.utils.numpy import rmse

import nengo_loihi

# nengo.Ensemble.max_rates.default = nengo.dists.Uniform(100, 120)
# nengo.Ensemble.intercepts.default = nengo.dists.Uniform(-1, 0.5)

def go(n_neurons, solver=nengo.solvers.LstsqL2(),
[...]

PR #158, opened 2018-12-18
Binary spike encoding

arvoelke:
This is a work-in-progress / proof-of-concept that improves the accuracy of a single layer encode/decode by roughly 5-fold, which helps close the gap between the `nengo_loihi` emulator and the reference `nengo` simulator (see figure below). In particular, the benchmark from #155 sees an improvement from 5% error to about 1% error.

This idea is courtesy of @tcstewar who suggested that the spike generator could be implemented using a binary code where each spike represents a bit of information in the binary representation of the node's input vector. This uses `2*dims*k` spike generators to represent the input vector with `2**k` precision. There is no variability from the ideal PSC, because the synapse is `None` and the code is transmitted precisely every time-step. All of the error (on the encoding side) comes from quantizing the signed input values to `k` bits.

The trade-off is this requires `O(k)` spikes to be multicast to all input neurons every time-step, whereas the previous on/off encoding scheme was sparse in time (i.e., trading between spike density, dt, tau, and input frequency). To further explore this trade-off, the on/off code should be extended to support `k` heterogeneous spike trains (i.e., inflate the generator's total spike count by a factor of `k` to reduce variability).

binary-encoding

The implementation is a work-in-progress; I took the path-of-least-resistance in order to test this out as quickly as possible. At the very least this should help provide a starting template for what different encoding schemes might look like in code.

TODO:
 - Why is there a degradation in performance at `600 <= n_neurons <= 800` (note: 10 trials)? 
 - How should this be generalized to support user-configurable/pluggable spike generators?
 - Unify the encoding logic in the same class (currently split across files).
 - How to get the right scaling factor where it's needed?
 - Compare to on/off encoding replicated `k` times.
 - Unit test.
 - Regression test.

Figure code is from #155, modified as follows:

```python
data = defaultdict(list)
for trial in range(10):
    for simulator in (nengo_loihi.Simulator, nengo.Simulator):
        print("Trial #", trial)
        for n_neurons in np.arange(100, 1100, 100, dtype=int):
            data['Trial'].append(trial)
            data['Simulator'].append(simulator.__module__)
            data['# Neurons'].append(n_neurons)
            data['RMSE'].append(go(n_neurons, simulator=simulator))

df = DataFrame(data)
plt.figure(figsize=(14, 6))
sns.lineplot(data=df, x='# Neurons', y='RMSE', hue='Simulator')
plt.show()
```

### DIFF
--- a/nengo_loihi/builder.py
+++ b/nengo_loihi/builder.py
@@ -16,6 +16,7 @@ from nengo.utils.builder import default_n_eval_points
 import nengo.utils.numpy as npext
 
 from nengo_loihi import conv
+from nengo_loihi.loihi_api import ENC_BITS
 from nengo_loihi.loihi_cx import (
     ChipReceiveNeurons,
     ChipReceiveNode,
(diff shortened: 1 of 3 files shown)


arvoelke:
I refactored the code and added a unit test that demonstrates this works perfectly using the Nengo simulator. But sometimes things go wrong with the Loihi emulator and I can't figure out why. The improvement in accuracy seems to depend on some combination of neuron model, stimulus, and use of the defaults:
```python
nengo.Ensemble.max_rates.default = nengo.dists.Uniform(100, 120)
nengo.Ensemble.intercepts.default = nengo.dists.Uniform(-1, 0.5)
```

The most favourable condition I've found on this branch is a linear ramping input, with the above defaults, and the default neuron model.

linear

In this case `nengo_loihi` and `nengo` simulators are nearly equal in accuracy! The master branch is also close, but not as close (not shown). But for other conditions, things fall apart or become even worse than on master, and I'm finding it difficult to debug / isolate the possible sources of error. The mean values look right but there is often some pretty crazy variability.  For example, in the simulation below, the only difference from the above is that the input is now a constant 1.

wild_variability

And curiously, it is only so extreme on this branch. It's much less variable on master (shown below), which is counter-intuitive. I'm wondering if the voltage-probing trick could be a confounding factor?

master

arvoelke:
Ping. I think we should still consider this, as the spike generators are a significant source of error, and this reduces that error by a factor of 2^k for chosen k (see unit test that verifies this).

PR #159, opened 2018-12-21
Refactor files, emulator, api

hunse:
Replacement for #30. Currently based on #139, #124, #132.

The main file moving around has been done in the first commit. I tried to change as little as possible with the code (other than fixing imports), so if we do have to rebase this, it should be a simple matter of copying the changes to the right place.

TODO:
- [x] Refactor emulator
- [x] Integrate hardware api.py/allocator.py/builder.py. I'd be nice to perhaps have some logical distinctions here, but I'm not sure if the current divisions make sense.
- [x] Renaming things (e.g. all the Cx classes lose the Cx)
- [x] Add *Group classes (e.g. AxonGroup, SynapseGroup), and split up the `discretize` function
- [x] Some minor changes to simulator.py (e.g. using iteritems). Do we want to worry about `iteritems`? I feel like the performance benefit is very very minimal unless we've got huge dictionaries, and of course the benefit would only be for Python 2 (Python 3 is already fast).
- [x] Find a good place for `test_model.py`. It currently only has one test (the noise test) in it. This test also needs to be fixed so that it asserts something (currently only a visual test).
- [x] Test everything off-chip and on-chip

Discussion:
- When do we want to do validation for chip constraints? Previously, we had done this all together when the Model gets passed to the emulator or hardware builder. #30 started to do some of this when we add objects (e.g. Synapses) to their respective containers (e.g. SynapseGroup).
- #30 introduced a CoreGroup, which would house a CompartmentGroup along with associated AxonGroup, SynapseGroup, and ProbeGroup. I haven't done that yet for this PR, instead opting to keep AxonGroup, SynapseGroup, and ProbeGroup as attributes of CompartmentGroup. This fits better with my mental model, since a CompartmentGroup can potentially span multiple cores (though we don't support this yet), and the Synapses, Axons, and Probes all exist in relation to particular compartments.


hunse:
I was just thinking about how this structure might change if/when we want to have multi-compartment neurons. In this case, maybe we want a `NeuronGroup` class where you can specify a certain number of neurons and a neuron model. The neuron model would define how the neuron is created using one or more compartments. The `NeuronGroup` could then be like `CoreGroup` in #30, and own a `CompartmentGroup`, `AxonGroup`, `SynapseGroup` and `ProbeGroup`. Does this structure make sense?

tbekolay:
> When do we want to do validation for chip constraints?

I'm cool with deferring it until model build to reduce the number of changes here.

> CoreGroup vs NeuronGroup

`NeuronGroup` is better, `CoreGroup` was a name that I was like "hopefully I'll come up with something better later" because it's the wrong abstraction level anyway (those objects aren't yet tied to a chip location, really).

tbekolay:
We could name it `EnsembleGroup`, put it in `ensemble.py`, that's maybe more consistent for a few reasons.

nengo_loihi/hardware/simulator.py
```diff
@@ -0,0 +1,441 @@
...
+from nengo_loihi.compartments import CxGroup
+from nengo_loihi.discretize import scale_pes_errors
+from nengo_loihi.hardware.allocators import one_to_one_allocator
+from nengo_loihi.hardware.builder import build_board
+from nengo_loihi.probes import CxProbe
+
+logger = logging.getLogger(__name__)
+
+
+class LoihiSimulator(object):
```
drasmuss: Occurred to me that since we're separating these by submodule now, we could just have `hardware.Simulator` and `emulator.Simulator`.  Pro is that it makes the relation between all these different simulators more clear, but con is that if you import the simulator directly (e.g. `from emulator import Simulator`) then it isn't as obvious in the code what type of Simulator you're working with.  I'm on the fence, don't feel strongly either way.  But thought I'd raise the possibility in case anyone else has opinions.

nengo_loihi/inputs.py
```diff
@@ -0,0 +1,132 @@
...
+            self.errors[t] += x
+        else:
+            self.errors[t] = np.array(x)
+
+    def collect_errors(self):
+        for t, x in self.errors.items():
+            yield (self.target, t, x)
+
+
+class HostSendNode(nengo.Node):
```
drasmuss: This one isn't really an input, but it does seem like it should be grouped with these other objects.  Maybe `io_objects.py`?

nengo_loihi/emulator/emulator.py
```diff
@@ -40,137 +39,364 @@ class Emulator(object):
...
+        for _ in range(steps):
+            self.step()
+
+    def step(self):
+        """Advance the simulation by 1 step (``dt`` seconds)."""
+        self.t += 1
+        self.compartments.advance_input()
+        self.synapses.inject_current(
+            self.t, self.inputs, self.axons, self.compartments.spiked)
+        # self.compartments.update_input(self.synapses)
```
drasmuss: Leftover comment

nengo_loihi/emulator/emulator.py
```diff
@@ -40,137 +39,364 @@ class Emulator(object):
...
-            for group in self.groups])
-
-        if group_dtype == np.int32:
-            assert (self.scaleU == 1).all()
-            assert (self.scaleV == 1).all()
-            self.decayU_fn = (
-                lambda x, u: decay_int(x, self.decayU, offset=1) + u)
-            self.decayV_fn = lambda x, u: decay_int(x, self.decayV) + u
+            end_ix += group.n
+            # end_ix += group.n_compartments
```
drasmuss: Leftover comment

nengo_loihi/emulator/emulator.py
```diff
@@ -40,137 +39,364 @@ class Emulator(object):
...
+
+        self.n_compartments = end_ix
+
+    @property
+    def dtype(self):
+        return self.groups[0].vth.dtype
+
+    # @property
+    # def n_compartments(self):
+    #     return sum(group.n_compartments for group in self.groups)
```
drasmuss: Leftover comment

nengo_loihi/emulator/emulator.py
```diff
@@ -40,137 +39,364 @@ class Emulator(object):
...
+        self.probes.update(self.t, self.compartments)
 
-        # --- allocate group memory
-        group_dtype = self.groups[0].vth.dtype
-        assert group_dtype in (np.float32, np.int32)
+    def get_probe_output(self, probe):
+        return self.probes[probe]
+
+
+class GroupInfo(object):
```
drasmuss: Could you add docstrings for all the new state classes (GroupInfo, IterableState, CompartmentState, NoiseState, AxonState, SynapseState, ProbeState)?

nengo_loihi/compartments.py
```diff
@@ -105,45 +103,18 @@ def __init__(self, n, label=None, location='core'):
...
-            self.named_axons[name] = axons
+        self.axons.add(axons, name=name)
 
     def add_probe(self, probe):
-        """Add a Probe object to ensemble."""
-        if probe.target is None:
-            probe.target = self
-        assert probe.target is self
-        self.probes.append(probe)
+        self.probes.add(probe)
```
drasmuss: add_synapses/add_axons/add_probe seem a bit redundant now, could we just get rid of them and change the corresponding calls `model.add_axons(...)` -> `model.axons.add(...)`?

nengo_loihi/probes.py
```diff
@@ -13,3 +13,27 @@ def __init__(self, target=None, key=None, slice=None, weights=None,
...
+        self.probes = []
+
+    def __iter__(self):
+        return iter(self.probes)
+
+    def add(self, probe):
+        """Add a Probe object to ensemble."""
+        # if probe.target is None:
+        #     probe.target = self
+        # assert probe.target is self
```
drasmuss: Leftover comments

nengo_loihi/synapses.py
```diff
@@ -356,3 +363,113 @@ def validate(self):
...
+    def __init__(self):
+        self.synapses = []
+        self.named_synapses = {}
+
+    # def __contains__(self, obj):
+    #     return obj in (self.synapses if isinstance(obj, Synapses) else
+    #                    self.named_synapses)
+
+    # def __getitem__(self, key):
+    #     return self.named_synapses[key]
```
drasmuss: Leftover comments

nengo_loihi/axons.py
```diff
@@ -97,3 +99,31 @@ def validate(self):
                 for atom, axon_id in zip(self.cx_atoms, axon_ids):
                     n_populations = self.target.axon_populations(axon_id)
                     assert 0 <= atom < n_populations
+
+
+class AxonGroup(object):
```
drasmuss: Could you add docstrings for AxonGroup/SynapseGroup/ProbeGroup as well?

drasmuss:
Rebased this to master (saved the previous history in `refactor-files-saved`, in case I messed something up).

drasmuss:
Note: all emulator tests passing, but there is still one test failure on hardware (`test_emulator.py:test_uv_overflow[1000]`) that I'm trying to figure out.

drasmuss:
All tests passing now on hardware and emulator.

nengo_loihi/hardware/simulator.py
hunse: I'd vote to leave it as-is. When we use these, we often import both, so we'd need to rename things anyway or use full names.

nengo_loihi/synapses.py
hunse: I left this in to remind me to talk about it. I was originally going to use `__contains__` and `__getitem__` to look for and get values from `named_synapses`. I ended up making the more explicit `by_name` and `has_name`, instead. Does this make sense? Using `__contains__` and `__getitem__` is definitely slicker, but then they don't match `__iter__` (which iterates over all synapses, not just those in `named_synapses`).

nengo_loihi/axons.py
hunse: I've added short ones. If you can think of anything more to add, let me know.

nengo_loihi/emulator/emulator.py
hunse: I've added some docstrings. I didn't document attributes for `CompartmentGroup` and `NoiseGroup`, because there are a lot and they map pretty directly from the corresponding values in `Compartments`.

nengo_loihi/probes.py
hunse: So, I'm silly. Of course the assert doesn't work anymore, because the probe targets the NeuronGroup, not this ProbeGroup. To check that it targets the NeuronGroup, we'd have to pass that in to the ProbeGroup and store a handle there. Maybe checking this is too much hand-holding. There's lots of ways that someone could mess up when trying to put together a model, and we can't catch all of them. Also, this would be double-checking for devs, not front-end users, since all these objects are configured by the builder for front-end users.

The other option (and I wouldn't do this here, due to time requirements), is if Probes are always targeting the NeuronGroup that contains them, then we don't need a target parameter at all, we can just infer it when we build the probe.

drasmuss: Yeah I'm fine leaving the assertion out since it's only dev-facing.

nengo_loihi/synapses.py
drasmuss: Yeah that reasoning makes sense to me, seems like that could definitely lead to some sneaky bugs if `__contains__` and `__iter__` are referring to different collections.

nengo_loihi/compartments.py
hunse: So it turns out we were relying on this. So I've reverted the change, and put the (broken) changed branch at `refactor-files-addfunctions`.

hunse:
This is all rebased to master, and I think all comments are addressed. Hopefully all the tests pass.

nengo_loihi/builder/builder.py
```diff
@@ -0,0 +1,177 @@
...
+    def validate(self):
+        if len(self.groups) == 0:
+            raise BuildError("No neurons marked for execution on-chip. "
+                             "Please mark some ensembles as on-chip.")
+
+        for group in self.groups:
+            group.validate()
+
+
+class Builder(object):
```
tbekolay: Since this essentially is the same as the Nengo builder but using our registered functions, in #30 I essentially cloned the class by making this a mostly empty subclass (see https://github.com/nengo/nengo-loihi/pull/30/commits/ff01e4da120919039dcbbb9ac9341d310bac1b08#diff-75a6663f57ab43eda3b594e818e01e0dR88). Did something change in here such that it's not a straight copy of the Nengo builder?

nengo_loihi/builder/decode_neurons.py
```diff
@@ -0,0 +1,311 @@
...
+            represented by these neurons. Typically, the inputs will be neurons
+            belonging to an Ensemble, and these weights will be decoders.
+        group_label : string (Default: None)
+            Optional label for the NeuronGroup.
+        syn_label : string (Default: None)
+            Optional label for the Synapses.
+
+        Returns
+        -------
+        cx : CompartmentGroup
```
tbekolay: Should be a `NeuronGroup` correct? Also since we're generally getting rid of the `cx` name it might be more clear to call this variable something else.

nengo_loihi/emulator/emulator.py
```diff
@@ -0,0 +1,666 @@
...
+    n_compartments : int
+        The total number of neuron compartments (given by GroupInfo).
+    slices : dict of {item: slice}
+        Maps an item to the ``group_info.slice`` for the group it belongs to.
+    strict : bool (Default: True)
+        Whether "undesired" chip effects (ex. overflow) raise errors (``True``)
+        or whether they only raise warnings (``False``).
+    """
+
+    @staticmethod
```
tbekolay: Should go after `__init__`

nengo_loihi/hardware/api.py
```diff
@@ -0,0 +1,370 @@
+from __future__ import division
```
tbekolay: Yeah I'm not sure what to do about the division of things in the `hardware` folder... right now it seems like `builder.py` has build things, and `api.py` has things that the builder needs, so they should kind of be in one file, but that file would be too big so it should be split up. I'm not super sure about the name `api.py` though, who would be using this API? Maybe we could break it up into a file with `Board`, `Chip`, and `Core`, maybe named `physical.py` or `core.py` or something like that, then `profiles.py` for the profiles. Then I'm not sure about `LoihiSpikeInput`, maybe that lives in `builder.py` ?

nengo_loihi/hardware/builder.py
```diff
@@ -0,0 +1,497 @@
+from __future__ import division
+
+import logging
+
+import numpy as np
+
+from nengo.utils.stdlib import groupby
```
tbekolay: Group with numpy

nengo_loihi/inputs.py
tbekolay: Whether it's input or output depends on whether you're thinking from the host or chip perspective, so I think it makes sense to call it something that doesn't force you into one perspective or the other. `io_objects.py` sounds good to me :+1:

nengo_loihi/neurongroup.py
```diff
@@ -0,0 +1,64 @@
+from nengo_loihi.axons import AxonGroup
```
tbekolay: I think I would call this `neurons.py`, and rename the current `neurons.py` to `neurontypes.py` ... or just put them in the same file, it wouldn't be that long of a file, and I don't think it's that confusing for them both to be in there.

nengo_loihi/utils.py
```diff
@@ -0,0 +1,17 @@
+import numpy as np
```
tbekolay: I'm :-1: on a `utils.py` file. I would say `shift` should go in `discretize.py` and `Profile` either in `hardware/profile.py` or `synapses.py`. Neither location is really ideal, but the presence of a `utils.py` file means we'll get lazy and put anything and everything in here rather than deciding where it should go.

nengo_loihi/neurongroup.py
hunse: My problem with putting them together is that one is for nengo API neuron_types, and the other is for the nengo_loihi API. I'd prefer to keep those separate.

I would vote for a `neurontypes.py` for what is now `neurons.py` (is that better than `neuron_types.py`? is there precedent?). The main downside is that this now differs from core nengo, where the neuron types are in `neurons.py`.

nengo_loihi/builder/ensemble.py
```diff
@@ -0,0 +1,190 @@
...
+                "The specified intercepts for %s lead to neurons with "
+                "negative or non-finite gain. Please adjust the intercepts so "
+                "that all gains are positive. For most neuron types (e.g., "
+                "LIF neurons) this is achieved by reducing the maximum "
+                "intercept value to below 1." % ens)
+
+    return gain, bias, max_rates, intercepts
+
+
+BuiltEnsemble = collections.namedtuple(
```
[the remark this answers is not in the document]
hunse: Since #30, we have modified a few of those functions, mainly to use Loihi tuning curves for the activations when solving for decoders, I think. So I think that's why I skipped it. But many of them can probably be imported, so we should go through and check.

hunse:
For CompartmentGroup vs NeuronGroup: a CompartmentGroup is a member of a NeuronGroup. Right now, the number of neurons is the same as the number of compartments, but the idea is that when we have multi-compartment neurons, you'd create a neuron group and say e.g. I want 100 neurons of type AlphaNeuron. If AlphaNeuron has three compartments, then the CompartmentGroup will have 300 compartments that implement the 100 neurons.

For some reason, I can't "reply" to some of the inline comments, so I will here. Re: renaming LoihiSimulator: The nice thing about `Emulator` is that it makes it clear both where the thing is (not on hardware) and what it does (emulates the hardware). My problem with just `Hardware` or `Loihi` is both of those say where it is, but they don't say what it does (simulates things on hardware/Loihi). I don't think it's the end of the world, so if we're going with one of the two, I guess `Hardware`? (I like the name `hardware` for the folder, and I agree having them be the same is nice. Having a `loihi` folder seems a bit confusing, because everything is Loihi-related in this project, but again, I'm not too opinionated on that.)

nengo_loihi/compartments.py
```diff
@@ -0,0 +1,413 @@
...
+    probes : list of CxProbe
+        CxProbes recording information from these compartments.
+    location : {"core", "cpu"}
+        Whether these compartments are on a Loihi core
+        or handled by the Loihi x86 processor (CPU).
+    """
+    # threshold at which U/V scaling is allowed
+    DECAY_SCALE_TH = 0.5 / 2**12  # half of one decay scaling unit
+
+    def __init__(self, n, label=None, location='core'):
```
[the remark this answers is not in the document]
hunse: This happens, just later.

nengo_loihi/emulator/emulator.py
```diff
@@ -0,0 +1,472 @@
+from __future__ import division
+
+# import collections
+import logging
+import warnings
+
+import numpy as np
+# import nengo
+from nengo.exceptions import SimulationError
+# from nengo.utils.compat import is_integer, is_iterable
```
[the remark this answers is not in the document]
hunse: I had removed it, just not in the first commit. Now I've removed it right away!

nengo_loihi/hardware/api.py
hunse: `nxsdk_objects.py`

nengo_loihi/hardware/simulator.py
hunse: `EmulatorInterface` and `HardwareInterface`

tbekolay:
We discussed a few ways to organize files yesterday, all with some pros/cons. It's easier to compare the options if we can see the final source tree, so here are those options.

# Option 1: Loihi-oriented :smile:

Couldn't think of a good name, but this is the current state of this PR. File names are roughly based on class names, which are tied to how the chip works.

In this and subsequent file listings, I've omitted test files and `__init__.py` files.

Here I've annotated with the # of lines of code, as we know this from the current branch. In subsequent options, I'll estimate the number of lines of code given what I assume will go in each file.

```
nengo_loihi
├── builder
│   ├── builder.py (190)
│   ├── connection.py (331)
│   ├── decode_neurons.py (311)
│   ├── ensemble.py (161)
│   ├── node.py (19)
│   └── probe.py (148)
├── emulator
│   └── emulator.py (667)
├── hardware
│   ├── allocators.py (128)
│   ├── builder.py (497)
│   ├── nxsdk_objects.py (371)
│   ├── nxsdk_shim.py (33)
│   └── simulator.py (431)
├── axons.py (139)
├── compartments.py (271)
├── config.py (46)
├── conv.py (715)
├── discretize.py (186)
├── io_objects.py (131)
├── neurongroup.py (65)
├── neurons.py (199)
├── passthrough.py (274)
├── probes.py (47)
├── simulator.py (753)
├── splitter.py (470)
├── synapses.py (500)
└── version.py (15)
```

# Option 2: Nengo-oriented :tada:

This is roughly what I was thinking about as a possible Nengo-oriented file structure when I was reviewing the PR last week.

```
nengo_loihi
├── builder
│   ├── builder.py (190)
│   ├── connection.py (331)
│   ├── decode_neurons.py (311)
│   ├── ensemble.py (161)
│   ├── node.py (19)
│   └── probe.py (148)
├── emulator
│   └── emulator.py (667)
├── hardware
│   ├── allocators.py (128)
│   ├── builder.py (497)
│   ├── nxsdk_objects.py (371)
│   ├── nxsdk_shim.py (33)
│   └── simulator.py (431)
├── connection.py (639 = 139[axons.py]+500[synapses.py])
├── config.py (46)
├── conv.py (715)
├── discretize.py (186)
├── esnemble.py (336 = 271[compartments.py]+65[neurongroup.py])
├── io_objects.py (131)
├── neurons.py (199)
├── passthrough.py (274)
├── probe.py (47)
├── simulator.py (753)
├── splitter.py (470)
└── version.py (15)
```

# Option 3: File suffixes, Nengo-oriented :heart:

This could also be prefixes, but I think suffixes are preferred. This is similar to how NengoDL is organized. Here, I combine some of the top-level files into the `_builders.py` files. This makes some of them quite long (notably `connection_builders.py`) but I think that once we put them together, we might be able to see some opportunities to refactor further, e.g. to put learning rule related things from `connection_builders.py` into `learning_rule_builders.py`.

```
nengo_loihi
├── emulator
│   └── emulator.py (667)
├── hardware
│   ├── allocators.py (128)
│   ├── builder.py (497)
│   ├── nxsdk_objects.py (371)
│   ├── nxsdk_shim.py (33)
│   └── simulator.py (431)
├── builder.py (190)
├── connection_builders.py (970 = 331+500[synapses.py]+139[axons.py])
├── config.py (46)
├── conv.py (715)
├── decode_neurons.py (311)
├── discretize.py (186)
├── ensemble_builders.py (497 = 161+271[compartments.py]+65[neurongroup.py])
├── io_objects.py (131)
├── neurons.py (199)
├── node_builders.py (19)
├── passthrough.py (274)
├── probe_builders.py (195 = 148+47[probes.py])
├── simulator.py (753)
├── splitter.py (470)
└── version.py (15)
```

There might also be an option 3b or 4 that is a variant of this option that keeps `axons.py`, but `synapses.py` is a bit tricky because in this case we would expect `synapses.py` to contain `Synapse` subclasses.

---

Vote with emoji reactions if this all makes sense, otherwise comment away!

hunse:
Option 5: Like option 1, but moves some files into an `api` folder (the files with classes that describe the mid-level Loihi model). I could also see this folder being called `loihi` or `model`. This means that none of the files in the root directory share names with `nengo` files but deal with different objects.

```
nengo_loihi
├── api
│   ├── axons.py (139)
│   ├── compartments.py (271)
│   ├── decode_neurons.py (311)
│   ├── neurongroup.py (65)
│   ├── probes.py (47)
│   └── synapses.py (500)
├── builder
│   ├── builder.py (190)
│   ├── connection.py (331)
│   ├── decode_neurons.py (311)
│   ├── ensemble.py (161)
│   ├── node.py (19)
│   └── probe.py (148)
├── emulator
│   └── emulator.py (667)
├── hardware
│   ├── allocators.py (128)
│   ├── builder.py (497)
│   ├── nxsdk_objects.py (371)
│   ├── nxsdk_shim.py (33)
│   └── simulator.py (431)
├── config.py (46)
├── conv.py (715)
├── discretize.py (186)
├── io_objects.py (131)
├── neurons.py (199)
├── passthrough.py (274)
├── simulator.py (753)
├── splitter.py (470)
└── version.py (15)
```

tbekolay:
The thing that I find confusing about Option 5 is what differentiates things in the `api` or `model` folder from things in the root folder. Why do `decode_neurons.py` go there, but not `io_objects.py` or `discretize.py` ? And if we're calling it the API or the underlying model, who are we expecting to use it? I see the user-facing API as being mostly `nengo_loihi.Simulator` and the things in `config.py` and `neurons.py`.

Since we have two for option 3, are there specific things that you don't like about it that we can change @hunse?

hunse:
I too was struggling with what exactly goes in the `api` folder. Since it's called `api`, I was thinking it's all files that hold objects associated with the mid-level Loihi model. `discretize.py` has many functions associated with this API, but also has functions used elsewhere, which is why I left it out. `io_objects.py` are mostly based off Nengo objects and used in `splitter.py`, so I kept it out.

It's the Loihi API that developers use to write models for the chip. Not all APIs have to be user-facing, in my opinion, but I can understand how just calling it `api` implies this. I'd be fine with `loihi_api`, or just `loihi`.

The biggest thing I don't like about option 3 is it hides this mid-level Loihi API (or whatever you want to call it). To me, this is the core around which `nengo_loihi` is built, and for anybody wishing to do considerable work on the project, I think it's the first thing that they should understand. To me, this API encapsulates how the chip is organized, without getting into the nitty-gritty. I feel like option 3 hides it and makes it more difficult to understand, which may seem better at first (since it hides complexity) but in the long-term is worse for somebody understanding the project. 

I also don't like that option 3 doesn't have a builder folder. This differs from `nengo` and is unexpected, and also makes for unwieldy filenames IMO.

It's also just different from the way I naturally group things. I tend to group objects at the same level of a hierarchy together, even if they're serving different functions. Option 3 groups things more functionally (e.g. everything related to connections together) grouping across levels in the hierarchy. There's no clear right or wrong, it's just not the way I tend to group things in my mind.

tbekolay:
Comments from @drasmuss on 5bf9d45, 32df40e, 1741123:

> I had to make a couple changes to avoid cyclic imports:
> 1. Moved `builder.Model` to `simulator.Model`.  `builder.Builder` gets imported all over the place (wherever we need to register a build function). But `builder.Model` has a lot of ties to other parts of the code base (decode neurons, neurongroups, spikeinputs). So if we tie `Builder` and `Model` together you end up spreading those `Model` dependencies all over the codebase. Separating them allows `builder.Builder` to be a nice lightweight import. I'm not commited to putting `Model` in `simulator`, but that seemed like a reasonable place :shrug:
> 2. Put `Axons` and `Synapses` in `ensemble_builder.py` instead of `connection_builder.py`.  `NeuronGroup`/`CompartmentGroup`/`AxonGroup`/`SynapseGroup` are closely tied together (e.g. a neurongroup includes compartmentgroups, axongroups, and synapsegroups). So to build an ensemble we need access to all of those things (because we need to build NeuronGroups). But to build connections, we also need access to NeuronGroups, for the specific case of building probe connections (where we need to create a post object to connect to and then probe). So those two can't mutually import from one another, we need all the `NeuronGroup` cluster in one location.
> 
> We could also resolve 2 by moving the `NeuronGroup` cluster to its own file (which connection_builders` and `ensemble_builders` would both import from).
>
> Or, we could redo the connection builder so that it doesn't require `NeuronGroup`. as mentioned, this is just for that specific connection-probe case. so maybe that could be done as part of the probe builder? I don't know enough about that build process to say how feasible that would be.
>
> Or, maybe having all those things in `ensemble_builder` is fine.

After thinking about this for a while, I implemented something like "move the `NeuronGroup` cluster to its own file" after updating my mental model a bit. Hopefully this mental model is a bit closer to what @hunse has in his head.

The key things I realized:

1. The first two classes a backend dev should look at first are `NeuronGroup` and `SpikeInput` as everything else stems from them.
2. The builder converts Nengo models to `NeuronGroup`s and `SpikeInput`s.

This might sound dead obvious, but this was not obvious to me until I thought about it last night, despite doing the first round of this refactoring in #30 and looking at the code quite a bit. Once I realized these things, I mulled around a few ideas about how to emphasize `NeuronGroup` and `SpikeInput` more, like making `NeuronGroup` a subclass of `nengo.Network` but instead of ensembles and connections it has compartments, axons, and synapses, but you get the nice context manager syntax and whatnot. That could still happen, but in the end I went with something that would require fewer changes to the code but still have the desired effect in 96b3c20. Specifically I:

1. Renamed `NeuronGroup` to `LoihiSegment`.

   I switched my way of thinking about `NeuronGroup` from "this is a related group of compartments, axons, synapses, and probes" to "this is a defined chunk of the Loihi system, be it emulator or hardware." Now, in my mind, I see the Loihi board as being a bunch of non-overlapping segments, like in an [image segmentation problem](https://cdn-images-1.medium.com/max/2000/1*hf6J8vsX7gmHhZnbPa4y9g.png). The `LoihiSegment` contains all the Loihi data structures defining that segment, so compartments, axons, synapses, and probes. The segments could span across multiples cores / chips / boards, and it's the job of the allocator to split / merge segments such that you end up with a bunch of core-specific segments.

   I'm not super wedded to the name `LoihiSegment`, but I like having `Loihi` at the start so that it is clear that it's a primary abstraction for dealing with Loihi. The `Segment` part is a bit harder to justify; I was originally thinking `Slice` (i.e., a slice of the Loihi system) but that's already a name with lots of meaning in Python/NumPy. Segment has the advantage of not being attached to any other meanings right now, and it has a nice analogy to image segmentation, which could be a nice way to explain how things are organized in our dev documentation when we get around to it. Other names I considered were `LoihiRegion`, `LoihiPiece` and a few others.

2. Made a `LoihiInput` base class for `SpikeInput` and moved those things to `inputs.py`.

   As the other primary abstraction that people should know, it should be prefaced with `Loihi`. If spike inputs were the only possibility then I would just rename it, but I was thinking that one other way to provide constant input is to make an ensemble with bias that will decode to the right value. We do something like this in several tests, and would be an efficient way to do static inputs -- it is, I believe, also possible to change bias values on the fly, so it could be a convenient thing for any precomputable input. In any case, there are other ways to do it, so it made more sense to split off a base class. If we're OK with all these things, then a TODO for later would be to move the spike generation stuff to `SpikeInput` so that the way in which you interact with the hypothetical `BiasInput` is the same as for `SpikeInput`, and can be put in the `LoihiInput` base class (but for now it's fine to assume everything is a `SpikeInput` I think).

3. Removed the `*Group` classes.

   These were just fancy lists that did a small number of extra things in certain situations. I put all those extra things in `LoihiSegment` and it's still a tiny 57-line class, so we can add many more fancy things without issue. And we should add more to the `LoihiSegment`, as we should expect most people will want to interact with it more than the underlying classes in most cases. For example, we currently would still do something like `segment.add_probe(Probe(...))`, but we easily change this to `segment.add(Probe(...))` or `segment.add_probe(...)`.

4. Moved the `LoihiSegment` and `LoihiInput` related classes to `segment.py` and `inputs.py` respectively.

   For `LoihiSegment` that's `Axons`, `Compartments`, `Synapses`, and `Probe`. For `LoihiInput`, that's `HostSendNode`, `HostReceiveNode`, `ChipReceiveNode`, `ChipReceiveNeurons`. 

   Sidenote: I think a TODO that we should do in this PR is to make a consistent decision about when we use plurals on class/module names and when we don't. I think that we can assume that all things represent multiple things, so while it's true that, for example, a segments contains multiple Loihi compartments, I feel like we can still call that class `Compartment` and not the current `Compartments` if we decide that's easier to remember (which it might be, hard to say).

5. Extracted the `discretize` and `validate` methods on various objects to `discretize.py` and `validate.py` respectively.

   I did this first for discretizing. There wasn't one big reason to do it, but a lot of little reasons, and all those together seemed enough for me.

   1. More in the previous organization than in this one, but a bunch of different files were importing the same functions from `discretize`, so grouping all of the discretization logic together keeps things in the same file and thus easier to reference, and cuts down on the number of import statements we need across all files.

   2. With the emphasis on `LoihiSegment`, it's important that the `segment.py` file is of reasonable length, as we essentially expect all devs to know that file pretty well. It felt to me like the discretization logic took up a lot of the literal and metaphorical space in that file, as you have to scroll past some intimidating looking math in order to figure out what a Synapse is and what it can, for example.

   3. Similarly, my mental model of Nengo Loihi is that discretization is an explicit, discrete step in the build process, so it felt like our file organization should reflect the fact that discretization is a big important thing and so if you want to learn about it, here's the place to do that.

   4. It feels like these functions are easier to test, even though I know that they are basically equivalently easy to test. I guess a related benefit would be that if we wanted to switch the logic such that we return new discretized segments instead of modifying them inplace, it feels more idiomatic to do this in a separate function than in a method inside the class.

   There might have been some other thoughts I had along the way, but these are the ones that stuck with me. Once I separated out the discretization methods, it seemed like all those reasons also applied to validation, so I separated them out too.

6. Moved `Model` back to `builder.py`.

   With the `LoihiSegment` and related classes moved to a separate file, the `Model` could go back to `builder.py` without issue.

After all these changes, there are no more circular imports (none of my decisions were changed on the basis of avoiding circular imports either) and IMO the files all have meaningful and descriptive names and are of reasonable length. Here's what the file tree with line numbers looks like now:

```
nengo_loihi
├── emulator
│   └── interface.py (673)
├── hardware
│   ├── allocators.py (129)
│   ├── builder.py (500)
│   ├── interface.py (431)
│   ├── nxsdk_objects.py (315)
│   ├── nxsdk_shim.py (33)
│   └── validate.py (68)
├── builder.py (176)
├── config.py (46)
├── connection_builders.py (341)
├── conv.py (710)
├── decode_neurons.py (310)
├── discretize.py (412)
├── ensemble_builders.py (147)
├── inputs.py (113)
├── neurons.py (200)
├── node_builders.py (16)
├── passthrough.py (275)
├── probe_builders.py (148)
├── segment.py (619)
├── simulator.py (754)
├── splitter.py (488)
├── validate.py (90)
└── version.py (15)
```

I feel more confident that we could onboard a new dev with this organization. There are still issues obviously which we should keep thinking about, but I feel like this is a good point to work from moving forward. Open to other suggestions too of course, let me know what you think @hunse and @drasmuss!

hunse:
Comments on numbered points above:

1. What about `LoihiGroup` instead of `LoihiSegment`? Objectively, they're about equally descriptive I think (for somebody new joining the project), and group is shorter and fits with our current nomenclature.
2. I like having the `LoihiInput` base class.
3. I think I'm fine with removing these group classes.
4. The division of `inputs.py` is a little arbitrary, IMO. `LoihiInput` and `SpikeInput` fit equally well in `segment.py` I think, but it also makes some sense to split them out especially if we add other inputs in the future. They are different than the other objects in that file, though. Also, why the name switch back to `inputs.py` from `io_objects.py`? Doesn't the reasoning that motivated that change still stand? (I mean, the one thing that ties them all together is that they're used as inputs to the chip, so for me, I think the name `inputs.py` works. But we did change it for a reason.) (I just realized that since `Model` is back in `builder.py`, then it actually wouldn't make as much sense to have `LoihiInput` in `segment.py`.)
5. I like the idea of splitting out the discretize functions, since in many ways they're like the builder, and that has its own space. One reason I made them part of the classes is that---unlike the builder---they actually modify the classes in place. It looks like that's mentioned in the `discretize_model` docstring, so that's probably enough for now.
6. Sounds good.

As for the aside about plurals vs not, I'd be fine switching everything to singular for the class names.

hunse:
One other thing: I'm still not convinced that having `connection_builders.py` etc. is better than having a `builder` folder. Core Nengo uses the latter, of course, so that's my frame of reference, here. Also, that's what the builder in this project is directly comparable to/replacing (it even uses a number of the build functions). And probably the frame of reference for people coming to the project.

`nengo_dl`, on the other hand, is a bit different. It makes full use of the Nengo builder to generate a signal/op model. So what it has in `connection_builders.py` etc. is just functions to translate signals/ops to Tensorflow signals/ops. This is a different kind of builder, which is partly why not having the builder folder works well there, I think.

tbekolay:
> 1. What about `LoihiGroup` instead of `LoihiSegment`? Objectively, they're about equally descriptive I think (for somebody new joining the project), and group is shorter and fits with our current nomenclature.

The reason I changed away from `Group` is because it doesn't really position `LoihiGroup` separate from `Axons`, `Synapses` etc, it feels like it's on the same level, where I want to be explicit that it is a higher level because it contains those things. So the way I did that was to instead position it more on the `Board`, `Core`, `Chip` level where you're thinking about the Loihi system, and the `LoihiSegment` is a somewhat more abstract (i.e., not actually tied to hardware) version of that, where we're saying "this is some chunk of a board/core/chip" instead of saying "this is a collection of axons, synapses, compartments and probes." It's kind of equivalent for us once we realize the group/segment is a higher level abstraction, but for newcomers it is not easy to realize that fact.

Also, `LoihiGroup` doesn't necessarily do a good job saying what it is a group of. Are `LoihiInput`s part of the group? There's no reason why they couldn't be. With `LoihiSegment`, since we're saying that this is a chunk of the hardware, then an input doesn't necessarily make sense there because an input is how we interface with the board, it isn't part of the board itself.

Not saying that these abstractions are perfect and 100% clear, just explaining my motivations for moving away from the `Group` name. Would be best to get another opinion on it.

> Also, why the name switch back to inputs.py from io_objects.py?

Yeah I realized after I switched it that it used to be `inputs`. Like I said earlier, the key thing was that a Nengo Loihi model consists _only_ of a set of `LoihiSegment`s and a set of `LoihiInput`s. We call it `inputs` in the model, so that should be reflected in the file structure. When I was cool with the switch to `io_objects.py` that wasn't my mental model.

The output part of it is a bit tricky since we don't really have explicit things to shuttle data off the chip. I think that we should, it would make the backend splitter / passthrough stuff more clear, but since we haven't done it yet putting it `inputs.py` seems OK because we will want to have that file around permanently, whereas the compromise `io_object.py` I would envision being split up into `inputs.py` and `outputs.py` eventually.

> I'd be fine switching everything to singular for the class names.

Me too, seems easier to remember that all are singular vs some singular, some plural.

> I'm still not convinced that having connection_builders.py etc. is better than having a builder folder

My concerns were more around the other files, I'm fine going back to a builder folder for the suffixed files :man_shrugging: So unless @drasmuss has strong objections let's do that.

hunse:
> Also, `LoihiGroup` doesn't necessarily do a good job saying what it is a group of. Are `LoihiInput`s part of the group? There's no reason why they couldn't be. With `LoihiSegment`, since we're saying that this is a chunk of the hardware, then an input doesn't necessarily make sense there because an input is how we interface with the board, it isn't part of the board itself.

I think you have many more associations with the word "segment" than I do.

What about `LoihiEnsemble`? Right now, Nengo Ensembles map one-to-one with this class. Their other main use is interneurons, which again are like Ensembles. This would highlight that they should have all the things you'd associate with an Ensemble (synapses, axons) but not inputs (which in Nengo associate with Nodes). So this is more of a Nengo-specific connotation, but would be a great connection for anyone familiar with Nengo, I think.

drasmuss:
I think moving things back into the builder folder makes sense, since now those files only contain the `build_*` functions rather than the midlevel objects as well.

No strong opinion on `LoihiGroup/Segment`, either one feels about equally meaningful to me.  The thing I worry about calling it `LoihiEnsemble` is that I'd then expect there to be a `LoihiConnection` as well,  whereas in the loihi mental model those things are kind of amalgamated.

tbekolay:
> If my thought process doesn't give you similar associations with segment

Also spending some time using it helps, it's hard to think about abstractly... using `group` to this point didn't seem to give us all a similar understanding

hunse:
> I mean, I didn't before I switched things and thought about it a bit, so it's not like a prior association I had. I tried to explain my reasoning so we'd all come to a shared understanding about things. If my thought process doesn't give you similar associations with segment, then we should think of something better, I just have trouble coming up with something else.

I think that your reasoning makes sense. All I'm saying is for somebody coming to the project, this isn't going to be immediately obvious. No matter what the name is, we're going to have to explain what it is, what it does, what's in it.

What about `LoihiBlock`? It basically means the same thing as `LoihiSegment`, but it's a little more concise, and it has a nice building-block feel to it (since these are essentially the building blocks that we use to make a Loihi model).

tbekolay:
I like `block`, feels more concrete like the board/core/chip, but also shorter :ok_hand:

Could you make those changes @hunse?

hunse:
Ok, I think I've made all the changes.

PR #161, opened 2019-01-04
CIFAR-10 on Loihi

hunse:
This has CIFAR-10 running on Loihi. It replaces #117.

The main feature addition here is allowing population spikes on inputs. This allows us to re-use axons in the three-channel CIFAR-10 input images, making it much easier to get things on the chip.

A lot of these commits can probably be squashed together. The script in the sandbox should also be moved to the sandbox repository.

### DIFF
--- a/nengo_loihi/hardware/nxsdk_objects.py
+++ b/nengo_loihi/hardware/nxsdk_objects.py
@@ -193,10 +193,42 @@ class Core(object):
 
 
 class LoihiSpikeInput(object):
-    class LoihiAxon(object):
-        __slots__ = ['chip_id', 'core_id', 'axon_id', 'atom']
+    """Stores information needed to send spikes to the actual chip.
 
-        def __init__(self, chip_id, core_id, axon_id, atom=0):
+    This acts as a bridge between a SpikeInput and the actual chip.
+    It maps positions in the spike input to actual locations on the chip.
+
+    Attributes
+    ----------
+    axon_map : {int: LoihiAxon}
+        Map from axon indices in the SpikeInput to LoihiAxons targeting
+        particular locations on the chip.
+    """
+
+    class LoihiAxon(object):
+        """Represents an axon going to the chip.
+
+        Parameters
+        ----------
+        axon_type : int
+            The population type of axon. 0 for discrete, 16 for pop16,
+            and 32 for pop32.
+        chip_id : int
+            The actual ID of the target chip on the board.
+        core_id : int
+            The actual ID of the target core on the board.
+        axon_id : int
+            The actual ID of the target axon on the board.
+        atom : int
+            The population index (atom), used if this axon sends population
+            spikes (i.e. axon_type != 0).
+        """
+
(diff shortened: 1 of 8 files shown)


hunse:
- [ ] TODO: remove xfails in `test_conv` tests that were waiting for the ability to send pop spikes to chip.

hunse:
So when I re-enable the `test_conv` tests, `test_conv.test_conv_connection` does not work for two-channel inputs. I need to look more into why this is. The rest of this could be reviewed (i.e. the notebook is not going to change much).

hunse:
The "docs" test is failing because it can't run the new notebook, because it needs the GPU TensorFlow. The CPU TensorFlow doesn't support having the channels first in conv, which is what I use here. We might be able to write our TensorFlow code to get around this, basically by explicitly transposing things if we realize we're on the CPU. Thoughts?

tbekolay:
Note: merging this is blocked by https://github.com/nengo/nengo-examples/pull/19 as we link to the example added in that PR here.

nengo_loihi/hardware/nxsdk_objects.py
```diff
@@ -294,9 +294,12 @@ def set_axons(self, board, n2board, spike_input):
...
                     self.axon_map.setdefault(input_idx, []).append(
                         self.LoihiAxon(
-                            axon_type=axon_type, chip_id=tchip.id,
-                            core_id=tcore.id, axon_id=taxon_id,
-                            atom=spike.atom))
+                            axon_type=axon_type,
+                            chip_id=tchip.id,
+                            core_id=tcore.id,
+                            axon_id=taxon_id,
+                            atom=spike.atom
```
hunse: If we're doing things this way, can we put a comma on the last line? Just one less change if we add more arguments later.

nengo_loihi/hardware/snips/nengo_io.c.template
```diff
@@ -17,19 +17,21 @@ void nengo_io(runState *s) {
...
 
-    int inChannel = getChannelID("nengo_io_h2c");
-    int outChannel = getChannelID("nengo_io_c2h");
+    int32_t axon_type;
+    int32_t axon_id;
+    int32_t atom;
     int32_t count[1];
     int32_t spike[SPIKE_SIZE];
-    int32_t error_info[2];
+    int32_t error_info[ERROR_INFO_SIZE];
```
hunse: This should be in the previous commit?

hunse:
All @tbekolay's additions/changes look good to me. Just two minor comments.

nengo_loihi/hardware/nxsdk_objects.py
tbekolay: Yes definitely, will add :+1:

nengo_loihi/hardware/snips/nengo_io.c.template
tbekolay: Correct, will fix! Thanks!

PR #162, opened 2019-01-04
Enable codecov status checks

drasmuss:
We won't require these checks to pass (as right now our CI can't cover the loihi hardware-specific code), but probably still useful to have the information readily displayed (and we would expect them to pass for things that don't touch the hardware).

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,10 @@
+codecov:
+  notify:
+    require_ci_to_pass: no
+
+coverage:
+  status:
+    patch:
+      default:
+        enabled: yes
+        target: 100%


tbekolay:
This in and of itself looks good to me, but I wonder if we should make a decision about setting organization-level codecov settings and repo-level codecov settings, since these overlap a lot with the NengoDL settings. We previously had some codecov configuration in Nengo core, but we [removed it](https://github.com/nengo/nengo/commit/dc85ab679e665f1644b59d8910058a41d25949e8) in lieu of [org level settings](https://codecov.io/account/gh/nengo/yaml), I wonder if we should put the common things in there so we don't have to configure it per-repo.

On the other hand, it's more explicit to have them in each repo, in which case we should make sure that we have minimal org-level settings and make a .codecov.yml again for Nengo core.

I think I'm in favor of modifying the org-level settings.

drasmuss:
I'm not super strong either way, but I think slightly in favour of having as much as we can of the configuration settings within the repo.  For example, when I was setting this up I forgot that there even were org-level settings, and spent a little while wondering why our behaviour was different than the defaults described in the codecov documentation.  It's nice to be able to just look at the `.codecov.yml` in the repo source and see everything you need to know about how it is configured, without having to go to a settings page somewhere else.

That being said, it does seem clunky to repeat the same settings in a bunch of different repos.  So yeah, I'm on the fence 🤷‍♂️ .

tbekolay:
Yeah that confusion probably takes more time than adding the settings to all repos. Currently the org level settings are

```
comment: off
coverage:
  status:
    project: off
```

I'm somewhat inclined to leave in `comment:off` because I find the codecov comments to be super ugly but remove the other entry, but I could also see removing it all.

drasmuss:
Yeah I like leaving `comment:off`.

tbekolay:
> Yeah I like leaving `comment:off`.

Sounds good, I made that change (see [org yaml history](https://codecov.io/account/gh/nengo/yaml/history)). Does that mean we can remove the `project:` section in this local .codecov.yml?

drasmuss:
Yeah, I'll do that

PR #169, opened 2019-01-17
Test probes

hunse:
This adds a test for neuron input (current) probes, which were not working correctly (due to negative values not being cast correctly, and integers not being filtered correctly). This also corrects both these issues.

TODO:
- [x] Add a test for voltage probes as well. The voltage should go negative, to test that we're dealing with negatives correctly. Do we currently allow `min_voltage < 0` for e.g. LIF neurons?

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -22,6 +22,14 @@ Release history
 0.5.0 (unreleased)
 ==================
 
+**Added**
+
+- Allow ``LIF.min_voltage`` to have effect. The exact minimum voltage on the
+  chip is highly affected by discritization (since the chip only allows
+  minimum voltages in powers of two), but this will at least provide something
+  in the ballpark.
+  (`#169 <https://github.com/nengo/nengo-loihi/pull/169>`__)
+
 **Changed**
 
 - PES learning in Nengo Loihi more closely matches learning in core Nengo.
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -49,6 +49,10 @@ Release history
   instead of ``dict``, which is not). This makes debugging easier and seeding
   consistent.
   (`#151 <https://github.com/nengo/nengo-loihi/pull/151>`_)
+- Probes that use snips on the chip (when running with ``precompute=False``)
+  now deal with negative values correctly.
+  (`#169 <https://github.com/nengo/nengo-loihi/pull/124>`_,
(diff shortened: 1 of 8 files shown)


nengo_loihi/loihi_interface.py
```diff
@@ -800,7 +800,7 @@ def _filter_probe(self, cx_probe, data):
             shape = data[0].shape
             synapse = cx_probe.synapse
             rng = None
-            step = (synapse.make_step(shape, shape, dt, rng, dtype=data.dtype)
+            step = (synapse.make_step(shape, shape, dt, rng, dtype=np.float32)
```
tbekolay: Any particular reason for `float32` over `float64` here? My default would be `float64` since that's what `float` and `np.float_` are.

nengo_loihi/loihi_interface.py
```diff
@@ -822,6 +822,7 @@ def get_probe_output(self, cx_probe):
         if cx_probe.use_snip:
             data = self._snip_probe_data[cx_probe]
             if cx_probe.synapse is not None:
+                data = np.asarray(data, dtype=np.float32)
```
tbekolay: Ditto float32 vs 64 here

nengo_loihi/loihi_cx.py
```diff
@@ -198,11 +198,12 @@ def _configure_filter(self, tau_s, dt):
                 "Current (U) scaling is required. Perhaps a synapse time "
                 "constant is too large in your model.")
 
-    def configure_lif(self, tau_rc=0.02, tau_ref=0.001, vth=1, dt=0.001):
+    def configure_lif(self, tau_rc=0.02, tau_ref=0.001, vth=1, dt=0.001,
+                      min_voltage=0):
```
tbekolay: Since we're touching this, can we add a docstring and document these parameters?

nengo_loihi/loihi_interface.py
hunse: Just to use less memory when we have a lot of probe data. Since the underlying values are all limited precision (since they come from Loihi), I figured that not much would be lost by using `float32` instead of `float64`.

hunse:
Addressed the requested changes. I've also added changelogs for both the probe fixes and the min_voltage.

hunse:
Fixups look good to me.

PR #176, opened 2019-01-30
Apply style checks to commit messages

drasmuss:
Uses `gitlint` to check for style on commit messages.  I added a bunch of example commits to verify that it catches the cases we want it to catch (see https://travis-ci.com/nengo/nengo-loihi/jobs/174091637#L1044).

I also changed our static check script back so that it runs all of the checks before exiting (rather than exiting immediately on error).  Particularly with the commit checking, it's helpful to see all of the commit message errors at once.

Also fixed an error in the `flake8` installation (unrelated to `gitlint`, just triggered by the new `flake8` release, but required to get the static checks passing again).

### DIFF
--- a/.ci/static.sh
+++ b/.ci/static.sh
@@ -8,6 +8,8 @@ NAME=$0
 COMMAND=$1
 
 if [[ "$COMMAND" == "install" ]]; then
+    # pip installs a more recent entrypoints version than conda
+    pip install entrypoints
     conda install --quiet jupyter
     pip install codespell flake8 pylint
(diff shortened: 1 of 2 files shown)


tbekolay:
One exception to these rules that comes to mind is that `Co-authored-by` lines might end up being too long because of long email addresses... I would propose that in those (somewhat uncommon) cases we just ignore the failure as it doesn't seem like there's a way to easily ignore specific lines in a commit message, though it might be something we could add ourselves if it becomes annoying.

PR #178, opened 2019-01-31
Unset bash settings after scripts

tbekolay:
This is mainly relevant if these scripts are sourced, which happens for `conda.sh` on TravisCI. Once a script is sourced, the rest of the bash session will print all lines, which litters the TravsisCI log with internal commands that we can ignore like `travis_nanoseconds`.

This PR also gives a more informative error message if no command is passed to the script.

### DIFF
--- a/.ci/common.sh
+++ b/.ci/common.sh
@@ -1,0 +1,8 @@
+#!/usr/bin/env bash
+# Small snippets common to all CI scripts.
+# All CI scripts should source this script.
+
+exe() {
+    echo "\$ $*";
+    "$@";
+}
--- a/.ci/conda.sh
+++ b/.ci/conda.sh
@@ -1,5 +1,10 @@
 #!/usr/bin/env bash
-set -e -v  # exit immediately on error, and print each line as it executes
+set -e  # exit immediately on error
+if [[ ! -e .ci/common.sh || ! -e nengo_loihi ]]; then
+    echo "Run this script from the root directory of this repository"
+    exit 1
+fi
+source .ci/common.sh
 
 # This script sets up the conda environment for all the other scripts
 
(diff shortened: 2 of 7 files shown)


hunse:
One thing that's a little yucky about printing lines in general is that it makes it hard to distinguish an error. For example, if I call `conda.sh` without a command, I get something like this:
```bash
    source activate test
elif [[ -z "$COMMAND" ]]; then
    echo "$NAME requires a command like 'install' or 'script'"
else
    echo "$NAME does not define $COMMAND"
fi
.ci/conda.sh requires a command like 'install' or 'script'

set +e +v  # reset options in case this is sourced
```
It's kind of hard to notice the error in there with all the other stuff that's printed.

The easiest solution I can think of is to just put "ERROR" at the start of these error messages, so that you get something like this:
```bash
    source activate test
elif [[ -z "$COMMAND" ]]; then
    echo "ERROR: $NAME requires a command like 'install' or 'script'"
else
    echo "ERROR: $NAME does not define $COMMAND"
fi
ERROR: .ci/conda.sh requires a command like 'install' or 'script'

set +e +v  # reset options in case this is sourced
```
Do others find this helpful?

tbekolay:
Personally I don't find `set -v` all that helpful in the first place. One alternative that I'd be happy to implement is to explicitly echo the commands we want to echo instead of the whole script. It reduces the noise while still giving the context of what is being run. I'll push a quick commit to show what that looks like.

tbekolay:
2bfd160f3854cc691eafc186bdb125ac463d4619 does the selective echoing and ensures that the scripts are run from the `nengo-loihi` root diretory (even if run from somewhere else)

1c6d06b87dc6d7687bf43459395d683b24075387 enables static checks on our CI scripts (I have these enabled locally and it informed some of the changes in these recent commits, so I figured it would be good to have enforced)

I am happy to pull one or both of these commits into separate PRs for future sprints, but making them fit into an hour with time to spare.

tbekolay:
TravisCI does weird things with `BASH_SOURCE` (https://github.com/travis-ci/travis-ci/issues/8392) so this fails. I'll try to fix and if it takes more than ten minutes I'll move to a separate PR

tbekolay:
Fixed by checking what directory you're running in rather than changing directories (e9e5a8ae19f357e1b957fa9836e6badaeed7b615). This will fail if you rename the root directory to something other than `nengo-loihi`, not sure if we really care about that situation since these files are mostly for TravisCI where the directory isn't renamed.

hunse:
I do like to run the `static.sh` script sometimes, and my root is named `nengo_loihi` instead of `nengo-loihi`. Why do we need this check? If it's required, is there a way to only do it only on TravisCI?

tbekolay:
> Why do we need this check? If it's required, is there a way to only do it only on TravisCI?

There are a few spots in these scripts that we assume you're in the root folder, so while the check isn't strictly required, some of the scripts will fail if you're not in that folder, so the check is to enforce a certain workflow in case the user isn't doing that. You can verify this now if you're in `master` by running `static.sh script` from within the `.ci` folder.

If people are renaming the root folder, then we have to do this another way. The check is relevant both locally and on TravisCI. I'll try to think of another way.

tbekolay:
@hunse suggested checking for a few things to exist as relative paths, so I did that in 4db7f3690217ddd01. This should work in 99.999% of cases, so I think we can safely round that to 100%.

PR #180, opened 2019-02-01
All split networks use same seeds as Nengo

hunse:
This is true for the host, host_pre, and chip networks.

This uses the `seed-loihi` branch of Nengo.


hunse:
This is being done in #70.

PR #183, opened 2019-02-04
Fix bug with scalar probe weights

tbekolay:
Scalar probe weights occur when probing spikes, which was not being handled correctly in the `EmulatorInterface`. This is used in the CIFAR-10 notebook.

### DIFF
--- a/nengo_loihi/emulator/interface.py
+++ b/nengo_loihi/emulator/interface.py
@@ -5,7 +5,7 @@ import logging
 import warnings
 
 from nengo.exceptions import SimulationError
-from nengo.utils.compat import iteritems, itervalues, range
+from nengo.utils.compat import is_array, is_number
 import numpy as np
 
 from nengo_loihi.discretize import (
@@ -230,7 +230,7 @@ class IterableState(object):
             warnings.warn(msg)
 
     def items(self):
-        return iteritems(self.slices)
+        return self.slices.items()
 
 
 class CompartmentState(IterableState):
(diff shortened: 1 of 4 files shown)


nengo_loihi/emulator/interface.py
```diff
@@ -5,7 +5,7 @@
 import warnings
 
 from nengo.exceptions import SimulationError
-from nengo.utils.compat import iteritems, itervalues, range
```
hunse: Why remove `range`?

tbekolay: In Python 2, we do this import so that `range` is `xrange`, which gives an iterator instead of generating the full list, which can be expensive. In Python 3, `range` is already an iterator, and there is no `xrange` in Python 3. Since Nengo Loihi is Python 3 only, we can just use `range`.

tbekolay: As for practically why I did it, I was adding some imports there while working on this, and with those new imports it broke onto a new line, so I was like ... might as well just remove range and it'll stay on one line.

hunse: Can't we get rid of `iteritems` and `itervalues` then, too? Also, if we're making these changes, I'd do them throughout the repo.

tbekolay: Yes, I forget where but at one point I grepped and get rid of all of the `iteritems` but either missed some or some were reintroduced with rebasing and whatnot. I'll do that now.

tbekolay: Yeah the only ones remaining were in `emulator/interface.py` which was introduced after I did that; done in58c448fa63ce5c71692715f451e351c8514a8fd3.

PR #185, opened 2019-02-06
100% coverage

hunse:
Get 100% coverage in codecov.

- [x] Cover all BuildErrors and other exceptions
- [x] Remove unused code
- [x] Test code that should not fail in tests (e.g. chip connection/communication) using mocks
- [x] Cover any remaining uncovered lines, or mark as "no cover" (e.g. failed imports)


hunse:
OK, I think I've got everything covered, except for the few cases that we've explicitly talked about.

Currently, this is still a lot of commits. I've organized them into added tests first, then code removals, then other stuff. I've kept them separate for now (I'm thinking it will be easier to review that way), but they can mostly be squashed before merging (maybe squashing all the tests into one commit, and most of the removals into another). The two removals we should keep separate IMO are the first two in the order right now: e29050a and dfa4da5. The first removes the NIF neurons and replaces them with SpikingRectifiedLinear for our node encoding. The second gets rid of probes from LoihiInput, which we don't use right now, but may want to add back in in the future to be able to probe parts of host->chip connections.

I also think we should keep the "other stuff" mostly separate (i.e. things that are not tests or removals), since they're all fairly different changes and some have potential functional consequences.

This is also based off #140 right now (to fix the tests), so we should merge that first.

hunse:
One thing we haven't discussed is `conftest.py`. It technically has one uncovered line. I would recommend removing it from coverage entirely.

nengo_loihi/emulator/interface.py
```diff
@@ -353,11 +353,9 @@ def __init__(self, block_info):
...
 
         if self.dtype == np.int32:
-            if np.any(self.exp < 7):
-                warnings.warn("Noise amplitude falls below lower limit")
-                self.exp[self.exp < 7] = 7
-
-            self.mult = np.where(self.enabled, 2**(self.exp - 7), 0)
+            # TODO: if we could do this mult with shifts, it'd be faster, but
+            # numpy has no function taking a vector of positive/negative shifts
+            self.mult = np.where(self.enabled, 2.**(self.exp - 7), 0)
```
drasmuss: Is this being tested anywhere for exp < 7?

tbekolay: Why was the `self.exp[self.exp < 7] = 7` removed? I added it because it was happening all the time and crashes when it does.

hunse: Well it shouldn't crash. That would be a bug with the emulator. Can you reproduce the crash?

The main reason I removed it is that we were getting different behaviour between the emulator and chip on a test. I believe that the chip will allow a noise exponent down to 0. It could be that I'm wrong, there, though. Either way, any clipping or warning about values should be done in `discretize.py` and `validate.py`, so that it happens for both emulator and chip. If the emulator specifically is crashing, then we need to fix it for these values.

pytest.ini
```diff
@@ -1,5 +1,5 @@
 [pytest]
-addopts = -p nengo.tests.options --simulator nengo_loihi.Simulator --neurons nengo.LIF,nengo.SpikingRectifiedLinear
+addopts = -p nengo.tests.options --simulator nengo_loihi.Simulator --ref-simulator nengo_loihi.Simulator --neurons nengo.LIF,nengo.SpikingRectifiedLinear
```
hunse: Do we expect to pass all the RefSimulator tests? Originally, I think this was designed for backends that use the full Nengo builder (like nengo_dl and nengo_ocl). But maybe we're able to pass all these tests...

nengo_loihi/hardware/builder.py
```diff
@@ -126,8 +126,6 @@ def build_core(n2core, core):  # noqa: C901
             if synapse.learning:
                 numStdp += len(axons)
                 assert np.all(axons >= firstLearningIndex)
-            # else:
-            #     assert np.all(axons < firstLearningIndex)
```
hunse: I do want to remember to do this when we have better allocators. I'll find the allocator issue and add a note.

nengo_loihi/emulator/interface.py
hunse: My guess is that the bug was because we had `2**(self.exp - 7)` instead of `2.**(self.exp - 7)`, so it wasn't allowing for floating point multiplication as needed. That's been changed here, so the bug should be resolved. We should make sure it's well tested though.

pytest.ini
drasmuss: I did run the test suite to verify that everything passed.

nengo_loihi/emulator/interface.py
hunse: In general, we don't have good tests for the noise. The one test that we have doesn't assert anything. I'll see if I can fix it up, since that's something we've wanted to do anyway, and it shouldn't be too difficult.

nengo_loihi/tests/test_inputs.py
```diff
@@ -1,5 +1,6 @@
 import nengo
 from nengo.exceptions import SimulationError
+import numpy as np
```
hunse: Accidental?

nengo_loihi/emulator/interface.py
hunse: Took a bit more time than I expected, because it revealed a bug that I ended up fixing. It fell loosely into my extra hour :)

nengo_loihi/tests/test_simulator.py
```diff
@@ -412,38 +412,74 @@ def test_double_run(precompute, Simulator, seed, allclose):
     assert allclose(sim1.data[probe], sim0.data[probe])
 
 
-def test_simulator_noise(request, plt, seed):
+@pytest.mark.parametrize('exp', [-4.5, -3, -2])
```
drasmuss: Is there any particular significance to these noise exponents? Like are -2 and -3 testing different things?  And do you know what these translate to in the discretized range (specifically, if we're testing exponents <7 and >7?).

hunse: Yeah, I did print out what they translate to, it's 5, 10, and 13. Basically I chose one to be discretized under 7, one to match the value that we were typically using for noise for our interneurons (-2) when we used noisy ones, and one in between. If you use too large of a value (e.g. -1) then some of the neurons may start firing, which messes up the statistics. We'd probably be fine with just -4.5 and -2, if we want to save on a test.

drasmuss: Those values make sense to me then.  Maybe just add a quick comment noting what they translate to?

hunse:
I've added changelog entries. There were only four commits that I thought deserved them.

PR #187, opened 2019-02-08
Raise warning instead of error for progress bars

drasmuss:
The GUI automatically tries to run with `progress_bar=True`, so raising a warning instead of an error allows the GUI to run successfully.


hunse:
I just added a `match` to the warning test, as well. We should really try to use that, particularly when testing for warnings, but even for exceptions, too. Otherwise, any warning will satisfy the check, and we've got a lot of warnings.

PR #188, opened 2019-02-12
Working with NxSDK 0.8

hunse:
Changes to support new NxSDK 0.8


hunse:
Ok, this should all be working and backwards compatible. Ready for review.

I wanted to use the new `nxsdk_version` in `nxsdk_shim.py` at the start of `HardwareInterface`, but unfortunately that breaks our tests where we monkeypatch the version to test our errors/warnings.

PR #189, opened 2019-02-12
Clean up test output

hunse:
Currently, just removes debug and info log messages.


tbekolay:
Rebased this, it's a tiny but probably beneficial change (the line removed here is essentially dead) but looking through the previous diff, I realized that we still have a `pytest.ini` file around. So perhaps in this PR we should:

- [ ] Move everything currently in `pytest.ini` to `setup.cfg` by modifying `.nengobones.yml`. I believe right now the `tool:pytest` section in `setup.cfg` is being ignored.
- [ ] Squash the first commit in here into the commit moving pytest config stuff.

drasmuss:
Moved everything from `pytest.ini` to `setup.cfg`, and also added a few more warnings to `filterwarnings` while I was touching this.

PR #191, opened 2019-02-22
Only collect coverage for nengo_loihi

tbekolay:
This came up when testing some stuff from #185 locally. Not sure if we do this in our TravisCI build through a command line argument, but regardless this is useful when running coverage tests locally (otherwise the coverage report has way too much to read).


hunse:
Done in e24ec50 (#215).

PR #192, opened 2019-02-22
Release 0.6.0

tbekolay:
We had some empty plots again, so I fixed those in 1bf0cc6 and also noticed a test that we can probably re-enable now with the changes we've made. They pass on emulator, but wasn't sure about hardware, so I figured I'd make a PR and let TravisCI do that testing for me.

I included the two release commits here so that if this PR passes tests I can merge this in and then go back to the 0.6.0 release commit, push things to PyPI and tag it and all that once it's in master.

### DIFF
--- a/nengo_loihi/tests/test_connection.py
+++ b/nengo_loihi/tests/test_connection.py
@@ -489,12 +489,14 @@ def test_ens_decoded_on_host(precompute, allclose, Simulator, seed, plt):
 def test_n2n_on_host(precompute, allclose, Simulator, seed_ens, seed, plt):
     """Ensure that neuron to neuron connections work on and off chip."""
 
+    if not seed_ens and nengo.version.version_info <= (2, 8, 0):
+        plt.saveas = None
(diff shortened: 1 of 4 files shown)


tbekolay:
> If possible we should try to handle the empty-plots-on-xfail issue in the `plt` fixture, but this works for now.

It's possible I think, but there are situations (maybe the majority of them) in which you want to see what the plot looks like for fails and xfails. What might work, however, is to somehow detect if there are no axes objects defined, and if there aren't to not generate that plot.

nengo_loihi/tests/test_connection.py
```diff
@@ -489,6 +489,9 @@ def test_ens_decoded_on_host(precompute, allclose, Simulator, seed, plt):
 def test_n2n_on_host(precompute, allclose, Simulator, seed_ens, seed, plt):
     """Ensure that neuron to neuron connections work on and off chip."""
 
+    if nengo.version.version_info <= (2, 8, 0):
```
tbekolay: This should have the `plt.saveas = None` line here too

PR #193, opened 2019-02-22
Improve speed of test suite

drasmuss:
Two improvements to test time.  First, use pytest-xdist to run tests in parallel.  This has the most significant impact on the hardware tests, as we can run jobs on different boards simultaneously.  Second, uses TravisCI's caching functionality to reduce the amount of time we spend setting up the environment for each job.  This will be slower for a cache miss, but should hopefully improve things overall (depends on how often the environment changes); we should keep an eye on it to make sure that we aren't getting too many cache misses.

### DIFF
--- a/nengo_loihi/hardware/nxsdk_shim.py
+++ b/nengo_loihi/hardware/nxsdk_shim.py
@@ -1,6 +1,8 @@
 from distutils.version import LooseVersion
 import os
+import shutil
 import sys
+import tempfile
 
 try:
     import nxsdk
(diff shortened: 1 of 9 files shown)


nengo_loihi/hardware/nxsdk_shim.py
```diff
@@ -13,6 +15,27 @@
     def assert_nxsdk():
         pass
 
+    from nxsdk.graph import graph
```
hunse: This might fail with older nxsdk, since they moved a bunch of this stuff around in 0.8.

hunse:
Sweet! I just glanced at it, so this isn't an official review, but it looks good. Also seems fairly straightforward compared to what I feared, particularly when you started talking about having to patch nxsdk.

nengo_loihi/hardware/nxsdk_shim.py
drasmuss: I checked and the Graph class is in the same place in 0.7.5.  However, I didn't actually try running things in 0.7.5 to confirm it works.

tbekolay:
The now-merged #204 parallelized the hardware tests, but I don't believe the caching was included in that PR, so I'm leaving this open for now.

drasmuss:
https://github.com/nengo/nengo-loihi/pull/219 added the caching, so this has all been done now

PR #194, opened 2019-02-25
Actually get coverage to 100%

hunse:
This time we're really going to do it!


tbekolay:
Rebased.

hunse:
I'm worried that it's used in the CIFAR 10 net. I'd rather test it or leave it than remove it.

drasmuss:
Do you have a code snippet showing the expected behaviour that I could turn into a test?

PR #201, opened 2019-03-22
Nengo version check

hunse:
Check Nengo version for compatibility. 

- Error if Nengo version is too old
- Warn if Nengo version is newer than latest tested version

TODO:
- [x] I moved the error to `__init__.py`, since otherwise if you use an earlier nengo version you just get an `ImportError` because `SpikingRectifiedLinear` doesn't exist in Nengo. So we need to check and error early (before these imports). However, I'm not sure how to test this early error, since it happens in `__init__.py` so we can't use monkeypatch (at least not in any way I'm aware of).

### DIFF
--- a/nengo_loihi/version.py
+++ b/nengo_loihi/version.py
@@ -5,10 +5,41 @@ and conform to PEP440 (see https://www.python.org/dev/peps/pep-0440/).
 '.devN' will be added to the version unless the code base represents
 a release version. Release versions are git tagged with the version.
 """
+import warnings
+
+
+def check_nengo_version():
+    import nengo
+
+    if nengo.version.version_info < minimum_nengo_version_info:
+        raise ValueError(
+            "nengo-loihi does not support Nengo version %s. "
+            "Upgrade with 'pip install --upgrade --no-deps nengo'."
+            % nengo.__version__)
+    elif nengo.version.version_info > latest_nengo_version_info:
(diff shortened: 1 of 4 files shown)


nengo_loihi/version.py
```diff
@@ -5,10 +5,40 @@
...
+version = "{v}{dev}".format(v=info2string(version_info),
                             dev=('.dev%d' % dev) if dev is not None else '')
+
+# --- Nengo version compatibility
+# oldest nengo version we are compatible with (test on release)
+minimum_nengo_version_info = (2, 7, 0)
+minimum_nengo_version = info2string(minimum_nengo_version_info)
+
+# newest nengo version we are compatible with (update on release)
+latest_nengo_version_info = (2, 8, 0)
```
drasmuss: The way I do it in NengoDL is that for dev versions I give the nengo dev version compatibility (which would be 3.0.0 in this case).  Because we are testing that, so they should be compatible, and it's also a pain getting this warning all the time when you're using dev versions.  Then when you do a release, you downgrade this to the latest supported nengo release (e.g. 2.8.0 if we were to do a release now).

Then I have this test https://github.com/nengo/nengo-dl/blob/master/nengo_dl/tests/test_version.py#L16 to make sure that we don't forget to update the version when doing a release.

I'd suggest that we do the same for nengo-loihi.

hunse: I think that that makes sense.

Issue #163, opened 2019-01-08
Passing instance config parameters to builder


drasmuss:
One of the things we like to use the config system for is passing backend-specific parameters to the builder. E.g.

``` python
with nengo.Network() as net:
    a = nengo.Node([0])
    b = nengo.Ensemble(10, 1)
    c = nengo.Connection(a, b)
    net.config[c].my_param = True

....

<inside nengo_loihi.builder>
def build_connection(model, conn):
    ...
    if model.toplevel.config[conn].my_param:
        <do something>
    else:
        <do a different thing>
```

The problem is that `splitter.py/passthrough.py` can change the front-end objects in the model, so in general `conn` in `build_connection` is not equal to `c` in the model definition.  And then we can't look up the config parameters that were set on `c`.  It would be good to have some way to pass instance config attributes like that to the builder.

Note that this is part of a larger issue we've talked about, establishing a clear connection between the objects the user creates and the objects that exist post splitter/passthrough manipulations (e.g. https://github.com/nengo/nengo-loihi/issues/154).  Or the even larger issue of refactoring the build logic so that more of this happens in the backend rather than the frontend (so the inputs to the builder would be the user objects, e.g. `c`, and the manipulations would happen after that point).

Issue #154, opened 2018-12-10
sim.data does not contain chip <-> host connections


arvoelke:
Discovered while trying to find a work-around to #152.

It may be useful to access the `BuiltConnection` object corresponding to connections from chip to host, for example to obtain the decoding weights. However, such an object cannot be found in `sim.data`, and neither can the connection from host to chip:

```python
import nengo
import nengo_loihi

with nengo.Network() as model:
    u = nengo.Node(output=0)
    x = nengo.Ensemble(100, 1)
    y = nengo.Node(size_in=1)
    
    conn1 = nengo.Connection(u, x)
    conn2 = nengo.Connection(x, y)
    
with nengo_loihi.Simulator(model) as sim:
    sim.run(1.0)    

print(conn1 in sim.data)
print(conn2 in sim.data)
```

```
---------------------------------------------------------------------------
AssertionError                            Traceback (most recent call last)
<ipython-input-93-2bd3edf8322b> in <module>()
     13     sim.run(1.0)
     14 
---> 15 print(conn1 in sim.data)
     16 print(conn2 in sim.data)

~/anaconda3/envs/py36/lib/python3.6/_collections_abc.py in __contains__(self, key)
    664     def __contains__(self, key):
    665         try:
--> 666             self[key]
    667         except KeyError:
    668             return False

~/CTN/nengo-loihi/nengo_loihi/simulator.py in __getitem__(self, key)
     45                     target = fallback.raw
     46                     break
---> 47         assert key in target, "probed object not found"
     48 
     49         if (key not in self._cache

AssertionError: probed object not found
```

Relatedly, the above stack trace reveals another error: `__getitem__` is raising an `AssertionError`, instead of a `KeyError`, and so the `conn1 in sim.data` is raising an error instead of returning `False`. `conn2 in sim.data` also raises the same error for the same reason (not found).

Swapping the simulator for `nengo.Simulator` prints `True` twice, as expected.

tbekolay:
This is a known issue (https://github.com/nengo/nengo-loihi/blob/master/nengo_loihi/tests/test_simulator.py#L45).

[...]

Issue #145, opened 2018-12-06
Efficient passthrough nodes


hunse:
On some models (including some Nengo tests), our current code to remove passthrough nodes is failing. This is often because removing the passthrough node entirely turns a factored `(m, 1) x (1, n)` connection into a full `(m, n)` connection, and the latter takes much more memory on the chip.

I proposed replacing passthrough nodes with interneurons, since that would allow the connection to stay factored in such places. Here's the discussion from Slack:

@hunse 
> A point that Aaron raised before, and I think is a good one, is that we're not very explicit or transparent about our mapping from a Nengo network to the network we put (with interneurons) on the chip.

> Passthrough nodes are often intentionally used at bottlenecks, where we have many ensembles feeding in (e.g. to different dimensions, like in an EnsembleArray) and we want to group them together.

> What if rather than removing passthrough nodes, we replaced them with interneuron groups. Then we can keep the connections factored, and don't have the problem of an e.g. (m, 1) x (1, m) connection turning into (m, m) weights, as Dan pointed out above. It would mean adding a few more features so e.g. we have more control over the weights between interneurons and the post ensemble (right now, they're always the encoders). But I think in the long run, being more explicit about interneurons would help clean up our builder, make things clearer and more customizable for users, and help fix this problem.

@tcstewar 
> I definitely like that idea.  There might be some intelligent thing about figuring out what passthrough nodes to keep (there are times when we just use them for organization, rather than for bottlenecks, such as in the circular convolution network), which will be an interesting challenge, but something in that direction seems to me like it'd be needed...

@hunse
> Any thoughts on how to do this? One option would be to make it part of the splitter, and basically have one or more rules that look for situations where we need interneurons and puts in Ensembles to represent those interneurons. The advantage to this is that the output is also a Nengo network, so you could simulate it with the reference backend to see what the effects of interneurons are. I think this would help considerably with debugging, since right now we have issues (e.g. with integrator accuracy, with two inputs not adding like we expect) and we're not sure how much of them is interneurons versus other aspects of mapping to Loihi (like quantization).

[...]

Issue #118, opened 2018-10-10
Indicate which parts of a model are running on/off chip


drasmuss:
It would be nice if there was some clear indication which parts of a model are executing on the chip and which are running on the host.  Perhaps we could log some summary description of the model at the end of the build process that would have that information?  Or add a flag into the built params (which could be accessed though `sim.data[my_obj]`)?

tbekolay:
We do have `logger.debug` messages for each object created and moved as part of the splitting process, which has been very useful when tests fail through pytest (as the log messages are displayed by default there). In #112 there is also a `location` method on a new class that prints out where the object is (one of `host`, `chip` or `host_pre`). We could expose that method through the `Simulator` also.

PR #202, opened 2019-03-26
Splitter refactoring

### LINKED ISSUE #80: Refactor builder logic
Right now our build logic is split across two qualitatively different phases, the splitter and the builder.  This allowed us to get a prototype running quickly, but it's definitely a big awkward to work with.  For example, if you want to understand how the radius is handled for an Ensemble, maybe you need to look at `build_ensemble`, maybe you need to look at `split` (on a Connection object), depending not only on whether that ensemble is on/off chip, but whether the presynaptic objects connected to that Ensemble are on or off chip. 

I think we could simplify the build logic a lot by integrating all the builder logic into the regular build process.  Basically this means handling the splitting through the backend, rather than by rearranging the front-end network objects.  This would be a significant change and a longer-term TODO item, but worth doing I think (perhaps after the next release).

arvoelke:
Closes #80. Refactors the splitter logic for adding / removing connections to occur within the builder. This consolidates the connection logic into one place (`builder/connection.py`) so that backend-relevant logic and objects are manipulated at the same level of abstraction.

Overarching ideas:
 - The splitter returns "directives" (i.e., sets of instructions) to guide the builder.
 - The builder fills in all three "sub-models" (`self`, `host`, and `host_pre`) in parallel, by delegating the objects to the appropriate builder methods.

Solves the following bugs:
 - https://github.com/nengo/nengo-loihi/issues/205 - Splitter handles sliced probes.
 - https://github.com/nengo/nengo-loihi/issues/206 - Passthrough removal handles sliced probes.
 - https://github.com/nengo/nengo-loihi/issues/207 - Decoder cache is not globally disabled.
 - https://github.com/nengo/nengo-loihi/issues/208 - Improved chip -> chip learning signal message.
 - https://github.com/nengo/nengo-loihi/issues/209 - Improved chip -> chip error signal message.
 - https://github.com/nengo/nengo-loihi/issues/211 - Splitter won't mutate original network.

Random improvements:
 - ~~`split_pre_from_host` is no longer `O(n^2)`~~ (This was a misunderstanding on my part, fueled by the fact the variable named "queue" was actually a stack -- it is now actually a queue)
 - `__contains__` is no longer `O(n)`

Bugs that have been identified that are outside the scope of this PR:
 - https://github.com/nengo/nengo-loihi/issues/210 - Passthrough connections aren’t seeded, and the order is non-deterministic.
 - https://github.com/nengo/nengo-loihi/issues/212 - Passthrough removal is too aggressive.
 - https://github.com/nengo/nengo-loihi/issues/213 - Passthrough removal is broken.
 - https://github.com/nengo/nengo-loihi/issues/214 - `precompute=True` could be smarter.

Existing issues that should now be more readily achievable:
 - #163 - Passing instance config parameters to builder
 - #154 - `sim.data` does not contain chip <-> host connections
 - #145 - Efficient passthrough nodes
 - #118 - Indicate which parts of a model are running on/off chip
 - #4 - Expose `params` and handle splitter-created objects

Other future work:
 - Ability to pass in `network=None`.
 - Change `assert self.model.dt == dt` to an error and then unit test (https://github.com/nengo/nengo/issues/1517).

### DIFF
--- a/nengo_loihi/builder/builder.py
+++ b/nengo_loihi/builder/builder.py
@@ -1,7 +1,8 @@
 from collections import defaultdict, OrderedDict
 import logging
 
-from nengo import Network
+from nengo import Ensemble, Network, Node, Probe
+from nengo.builder import Model as NengoModel
 from nengo.builder.builder import Builder as NengoBuilder
 from nengo.builder.network import build_network
 from nengo.cache import NoDecoderCache
@@ -106,9 +107,22 @@ class Model:
         self.build_callback = None
         self.decoder_cache = NoDecoderCache()
 
+        # TODO: these models may not look/behave exactly the same as
+        # standard nengo models, because they don't have a toplevel network
+        # built into them or configs set
+        self.host_pre = NengoModel(
+            dt=float(dt), label="%s:host_pre, dt=%f" % (label, dt),
+        )
+        self.host = NengoModel(
+            dt=float(dt), label="%s:host, dt=%f" % (label, dt),
+        )
+
         # Objects created by the model for simulation on Loihi
         self.inputs = OrderedDict()
         self.blocks = OrderedDict()
+
+        # Will be filled in by the simulator __init__
+        self.split = None
 
         # Will be filled in by the network builder
         self.toplevel = None
@@ -145,8 +159,11 @@ class Model:
         # magnitude/weight resolution)
         self.pes_wgt_exp = 4
 
-        # Will be provided by Simulator
+        # Used to track interactions between host models
         self.chip2host_params = {}
+        self.chip2host_receivers = OrderedDict()
+        self.host2chip_senders = OrderedDict()
+        self.needs_sender = {}
 
     def __getstate__(self):
         raise NotImplementedError("Can't pickle nengo_loihi.builder.Model")
@@ -168,13 +185,38 @@ class Model:
         self.blocks[block] = len(self.blocks)
 
     def build(self, obj, *args, **kwargs):
-        built = self.builder.build(self, obj, *args, **kwargs)
+        # Don't build the objects marked as "to_remove" by PassthroughSplit
+        if obj in self.split.passthrough.to_remove:
+            return None
+
+        if not isinstance(obj, (Node, Ensemble, Probe)):
+            model = self
+        elif self.split.on_chip(obj):
+            model = self
+        else:
+            # Note: callbacks for the host_model will not be invoked
+            model = self.host_model(obj)
+
+            # done for compatibility with nengo<=2.8.0
+            # otherwise we could just copy over the initial
+            # seeding to all other models
+            model.seeds[obj] = self.seeds[obj]
+            model.seeded[obj] = self.seeded[obj]
+
+        built = model.builder.build(model, obj, *args, **kwargs)
         if self.build_callback is not None:
             self.build_callback(obj)
         return built
 
     def has_built(self, obj):
         return obj in self.params
+
+    def host_model(self, obj):
+        """Returns the Model corresponding to where obj should be built."""
+        if self.split.is_precomputable(obj):
+            return self.host_pre
+        else:
+            return self.host
 
 
 class Builder(NengoBuilder):
--- a/nengo_loihi/builder/builder.py
+++ b/nengo_loihi/builder/builder.py
@@ -111,10 +111,14 @@ class Model:
         # standard nengo models, because they don't have a toplevel network
         # built into them or configs set
         self.host_pre = NengoModel(
-            dt=float(dt), label="%s:host_pre, dt=%f" % (label, dt),
+            dt=float(dt),
+            label="%s:host_pre, dt=%f" % (label, dt),
+            decoder_cache=NoDecoderCache(),
         )
         self.host = NengoModel(
-            dt=float(dt), label="%s:host, dt=%f" % (label, dt),
+            dt=float(dt),
+            label="%s:host, dt=%f" % (label, dt),
+            decoder_cache=NoDecoderCache(),
         )
 
         # Objects created by the model for simulation on Loihi
--- a/nengo_loihi/builder/connection.py
+++ b/nengo_loihi/builder/connection.py
@@ -1,7 +1,8 @@
 import copy
+import logging
 
 import nengo
-from nengo import Ensemble, Connection, Node
+from nengo import Ensemble, Connection, Node, Probe as NengoProbe
 from nengo.builder.connection import (
     build_no_solver as _build_no_solver,
     BuiltConnection,
@@ -9,6 +10,7 @@ from nengo.builder.connection import (
     get_targets,
     multiply,
 )
+from nengo.connection import LearningRule
 from nengo.ensemble import Neurons
 from nengo.exceptions import BuildError, ValidationError
 from nengo.solvers import NoSolver, Solver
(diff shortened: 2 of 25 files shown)


arvoelke:
FYI, the doc build gave me:
```
/home/travis/build/nengo/nengo-loihi/docs/examples.rst.rst:33:broken link: https://github.com/nengo/nengo-examples/blob/master/loihi-dl/cifar10_convnet.ipynb (403 Client Error: Forbidden for url: https://github.com/nengo/nengo-examples/blob/master/loihi-dl/cifar10_convnet.ipynb)
```

nengo_loihi/builder/tests/test_connection.py
```diff
@@ -0,0 +1,67 @@
...
+
+    with pytest.raises(BuildError, match="Conv2D"):
+        with Simulator(net):
+            pass
+
+
+@pytest.mark.parametrize("pre_dims", [1, 3])
+@pytest.mark.parametrize("post_dims", [1, 3])
+@pytest.mark.parametrize("learn", [True, False])
+@pytest.mark.parametrize("use_solver", [True, False])
```
hunse: This is 16 tests. Do we need every possible combination here? They don't take long on the emulator, but I'm worried they could add more time than necessary on the chip.

nengo_loihi/simulator.py
```diff
@@ -306,64 +305,65 @@ def __init__(  # noqa: C901
...
-                # If there is no host and precompute=False, then all objects
-                # must be on the chip, which is precomputable in the sense that
-                # no communication has to happen with the host.
-                # We could warn about this, but we want to avoid people having
-                # to specify `precompute` unless they absolutely have to.
-                self.precompute = True
-
-            # Build the network into the model
-            self.model.build(network)
+        if network is None:
```
hunse: Why stop allowing `network == None`?

nengo_loihi/splitter.py
```diff
@@ -1,543 +1,141 @@
...
-    split_host_to_learning_rules(networks, conns)
-
-    # --- Step 4: place precomputable parts of host
-    if precompute:
-        split_pre_from_host(networks)
-
-    # --- Step 5: place probes
-    place_probes(networks)
+            self.passthrough_directive = PassthroughDirective(
+                set(), set(), set())
```
hunse: Seems a bit strange to import both `convert_passthroughs` and `PassthroughDirective`. I would pick one as the interface, and have it somehow able to return and empty directive in the case that `remove_passthrough == False`.

nengo_loihi/splitter.py
```diff
@@ -1,543 +1,141 @@
...
-        networks.remove(c)
+            self._host_precomputable_objects = set()
 
-    conns = networks.original.all_connections
-    for c in remove_conns:
-        conns.remove(c)
-    conns.extend(add_conns)
+    def _preclosure(self):  # noqa: C901
+        # performs a transitive closure on all host objects that
+        # send data to the chip
```
hunse: What does this mean? Also, I'd make this a brief docstring.

nengo_loihi/splitter.py
```diff
@@ -1,543 +1,141 @@
...
-            if conn in inputs[node_or_ens]:
-                obj = base_obj(conn.pre_obj)
-            elif conn in outputs[node_or_ens]:
-                obj = base_obj(conn.post_obj)
-
-            if (isinstance(obj, (Node, Ensemble))
-                    and networks.location(obj) == "host"):
-                if isinstance(obj, HostReceiveNode):
+        # initialize queue with the pre objects on
+        # host -> chip connections
```
hunse: Why are all `pre` objects on host->chip connections precomputable? Couldn't you have something like node->ensemble->node->ensemble->node (where nodes are on host and ensembles are on chip) where the middle node would be the `pre` of a host->chip connection but not be precomputable?

nengo_loihi/builder/node.py
```diff
@@ -1,7 +1,7 @@
 from nengo import Node
 
 from nengo_loihi.builder.builder import Builder
-from nengo_loihi.inputs import ChipReceiveNode, SpikeInput
+from nengo_loihi.builder.inputs import ChipReceiveNode, SpikeInput
```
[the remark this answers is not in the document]
arvoelke: Yeah I was thinking about separating out that file as well. Currently `SpikeInput` is the only one that is used from outside the `/builder/` directory. In particular, `hardware/builder.py` and `emulator/tests/test_interface.py` import `SpikeInput`, and the rest are contained within the `/builder/`.

```$ grep "from nengo_loihi.builder.inputs" . -r
./nengo_loihi/builder/node.py:from nengo_loihi.builder.inputs import ChipReceiveNode, SpikeInput
./nengo_loihi/builder/connection.py:from nengo_loihi.builder.inputs import (
./nengo_loihi/builder/builder.py:from nengo_loihi.builder.inputs import LoihiInput
./nengo_loihi/builder/tests/test_inputs.py:from nengo_loihi.builder.inputs import ChipReceiveNode, PESModulatoryTarget
./nengo_loihi/conv.py:from nengo_loihi.builder.inputs import ChipReceiveNeurons, LoihiInput
./nengo_loihi/tests/test_simulator.py:from nengo_loihi.builder.inputs import SpikeInput
./nengo_loihi/tests/test_block.py:from nengo_loihi.builder.inputs import SpikeInput
./nengo_loihi/hardware/builder.py:from nengo_loihi.builder.inputs import SpikeInput
./nengo_loihi/emulator/tests/test_interface.py:from nengo_loihi.builder.inputs import SpikeInput
```

One exception is `conv.py`, which imports `ChipReceiveNeurons` and `LoihiInput` to use within `build_conv2d_connection`. However, there is a TODO in `builder/connection.py::build_chip_connection` next to the call to `build_conv2d_connection` that says "TODO: integrate these into the same function" and so this is supposed to be in `/builder/` anyway.

What if I were to move `LoihiInput` and `SpikeInput` to `block.py`?

nengo_loihi/builder/tests/test_connection.py
arvoelke: This unit test was copied from the old splitter tests. Perhaps this should be added to a card to investigate/prune the unit tests for CI?

nengo_loihi/simulator.py
arvoelke: We had some discussion about this in nengo slack. This wasn't being used (or tested) anywhere, and---prior to the refactoring---didn't make a whole lot of sense, because, if we look at these lines in the original version:

https://github.com/nengo/nengo-loihi/blob/da2053ff5847fa8c68b4a4d30378c2cead1edb58/nengo_loihi/simulator.py#L313-L366

there are a large number of attributes added to `self` within this block that are required later. So a user would need to do a few very special things in order to get this to work, which might easily change. Exposing this kind of untested surface area for no good reason seemed like a bad idea to me. @tbekolay agreed that it made sense to disallow for now.

After this refactoring is complete, it could make sense to add back in. Currently I think the need for the attribute `self.model.splitter_directive` being filled in is really the only thing holding this back for now. I'd love to have a better way of organizing the way that model parameters are filled in within the simulator / builder.

nengo_loihi/splitter.py
arvoelke: Yeah it's also strange that I import `is_passthrough` and have to manually filter things before calling the function. But right now all this weirdness/friction is limited to 6 lines of code, and I have a feeling all of it will change when we redo the passthrough removal. Would it be better to define an `EmptyPassthroughDirective` within passthrough.py and then import+use that?

nengo_loihi/splitter.py
arvoelke: https://en.wikipedia.org/wiki/Transitive_closure#In_graph_theory

Will add docstring.

nengo_loihi/splitter.py
arvoelke: Great catch... I'm amazed that none of the tests caught this. I pushed a commit 0cbe29b which tests this and makes the fix (namely, using the right set of connections to form the adjacency lists).

hunse: This still appears to assume that all `pre` objects on host->chip connections are precomputable. This might be true in some cases, like in the test you added where the passthrough nodes can be removed, but what if those weren't passthroughs (i.e. if all nodes in that test computed functions)? I think in general, you can have networks where some `pre` objects of connections are not precomputable.

nengo_loihi/splitter.py
hunse: I don't think that would make much of a difference. That's a fair point that it's only 6 lines and might change, and there's not an obvious way to do it better.

nengo_loihi/splitter.py
arvoelke: Yep.. this is the same assumption that passthrough removal has always been making. If anything in the transitive closure of objects that send to host (after passthrough removal) also receives data from the host then a `BuildError` is raised. That is, the example that you suggested currently (and previously) raises an error. There is a unit test for this (`test_split_precompute_loop_error`). Note I didn't change the algorithm here. I just updated it to be a bit clearer IMO and more efficient (it used to be `O(n^2)`). I talked about this in slack but there was no interest in making this smarter at the moment. I've summarized this alongside a similar issue now in #214.

The difference that you initially pointed out stemmed from the fact that previously the passthrough removal would be physically applied to the `SplitNetworks` object before computing the closure. In this PR the list of connections weren't changed, so I added the above commit to make sure it's using the same list as it had been in the original splitter.

By the way we should remove "*Various improvements to passthrough removal, including not removing useful passthrough nodes.*" from the commit message. This was squashed from one of the other commits that was fixing a mistake I had previously made. There aren't any improvements to passthrough removal in this PR apart from making it `O(n)` instead of `O(n^2)`.

nengo_loihi/simulator.py
hunse: I agree, there's not a whole lot of demand for it and it's not well-tested. It is something that we allow in the reference backend. I think going forward, we need to discuss whether it's something we want to support, and if so, add a `nengo` test for it to make sure backends support it. I've made an issue nengo/nengo#1519.

nengo_loihi/builder/tests/test_connection.py
hunse: Dan already did some of this not too long ago. If this is from before, it's probably already fairly efficient. We can leave it for now, and perhaps in a few months we'll be revisiting tests again and can do some pruning.

nengo_loihi/builder/node.py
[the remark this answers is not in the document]
hunse: If we're leaving the builder-specific ones in `builder/`, then, `LoihiInput` and `SpikeInput` *could* go in `block.py`. I'm trying to predict whether we'll have more in the near future, or not, because that would be the main thing that would motivate having `inputs.py`. The main potential one I can see on the horizon is `DVSInput`.

I think I'd vote to keep `LoihiInput` and `SpikeInput` in `inputs.py`, and move the others to `builder/inputs.py`.

Also, we should move the conv connection builder to `builder/connection.py` I think.

nengo_loihi/splitter.py
hunse: Okay, maybe some comments in the code to that effect would be helpful. I can draft them.

That's good to know about that part of the commit message. I had wondered about that when I put it in, because I couldn't see where the stuff about "not removing useful passthrough nodes" happened, but I didn't want to get rid of it until I checked with you. As for "various improvements", I was referring to renaming "offchip" to "ignore", supporting `ObjView`, and returning a named tuple instead of a tuple. I don't actually see that the passthrough removal algorithm changes at all.

arvoelke: "I don't actually see that the passthrough removal algorithm changes at all." Which part are you referring to here? The `O(n^2) -> O(n)` part? This is just from avoiding the repeated list pops, which each take `O(n)` time, by instead advancing a `head` index.

tbekolay: List pops in Python are O(1) (see [time complexity](https://wiki.python.org/moin/TimeComplexity))

arvoelke: That's from an amortized analysis or average case analysis. Here we were always popping from the beginning of a list of size `O(n)` which is `O(n)` each pop. Quote from the same page:

> Internally, a list is represented as an array; the largest costs come from growing beyond the current allocation size (because everything must move), or from inserting or deleting somewhere near the beginning (because everything after that must move). If you need to add/remove at both ends, consider using a collections.deque instead.

Since lists are just arrays, `list.pop(0)` needs to shift the entire array each time.

tbekolay: I thought we were doing `queue.pop()` ? Which pops from the end of the list

arvoelke: Ah, you are right. The naming of the variable was a misnomer. Since we append to the last and pop from the last, it is not a queue -- it is a _stack_. It doesn't actually matter which one it is, but I was trying to keep things consistent with a queue. So either we can change it back to `pop` and rename it from `queue -> stack` or leave it as is. The ordering shouldn't matter, but should make it clear.

hunse: This isn't the passthrough removal algorithm, though, this is the splitter. That's all I was trying to say, is that I didn't see any changes to passthrough removal.

arvoelke: Apparently my brain has started interchanging the words "passthrough removal" and "splitter" as well as "chip" and "host". Ha ha .. :x You're right that I didn't change anything about passthrough removal other than fixing a bug that affected probed `ObjView`s.

arvoelke: Pushed commit 433b489 and updated issue #214 according to the above conversation about precomputing.

arvoelke:
I ran @tcstewar's benchmark notebook with:
```python
conditions = [
    'nengo',                                         # the normal nengo backend
    'nengo_loihi(precompute=False, target="loihi")', # using Loihi itself
    'nengo_loihi(precompute=True, target="loihi")',  # using Loihi itself with precompute
]
```
on both `master` and this branch.

On this branch, `precompute=False` and `precompute=True` gave identical accuracy (but with different speeds) across all benchmarks and so I added a simple unit test (45eced7) to verify this. The learning benchmarks aren't supposed to work with `precompute=True` (#214) and so I added a commit (433b489) to improve the error message.

On master, `precompute=False` and `precompute=True` gave different results for `['CircularConvolution', 'MatrixMultiply', 'SPASequence']`, which I presume has something to do with some interaction between `remove_passthroughs=True` and `precompute` with the seeding of the network in the old splitter code (also see issue #210).

arvoelke:
Pushed two more commits (1711044, 0ca11d8) to move `LoihiInput` and `SpikeInput` from `builder/inputs.py` back into `inputs.py` and move `build_conv2d_connection` from `conv.py` to `builder/connections.py`. This should now address all your comments @hunse.

arvoelke:
Added a commit (6c201e0) that increases test coverage since I touched the lines of some un-tested convolution code.

nengo_loihi/builder/connection.py
```diff
@@ -282,6 +282,93 @@ def build_host_to_learning_rule(model, conn):
     nengo_model.build(pre2send)
 
 
+def build_conv2d_connection(model, conn):
```
hunse: I would probably put it after `build_chip_connection`, since it's a subset of that, but not a big deal either way.

nengo_loihi/builder/builder.py
```diff
@@ -167,8 +189,38 @@ def add_block(self, block):
         assert block not in self.blocks
         self.blocks[block] = len(self.blocks)
 
+    def delegate(self, obj):
```
hunse: It would be helpful to have a docstring here.

nengo_loihi/builder/connection.py
```diff
@@ -1,24 +1,372 @@
+import copy
+import logging
+
 import nengo
-from nengo import Ensemble, Connection, Node
+from nengo import Ensemble, Connection, Node, Probe as NengoProbe
```
hunse: I would maybe just `import nengo` and use `nengo.` for all of these.

hunse:
Just a few minor comments. Otherwise everything LGTM. Thanks for doing all those tests against the benchmarks.

arvoelke:
I added a couple very minor commits for coverage / sanity. I also rebased onto master again to pick up the testing fixes from the current/rates sync-up.

> 5694c27 Is Split a good name? Or should we stick with Splitter, SplitterDirective, or some other option?
15890be Similarly, is PassthroughSplit a good name?
ea6d52e Should this be a separate PR or is it OK to include here?

LGTM!

> 16e322a Is get_seed something we want built into Model? And if so, do we do it now or make a separate issue for it?

I think it's fine as is. A bunch of the weirdness with seeding can be resolved when we eventually drop support for nengo<3. This could end up simplifying a bunch of this code.

> f64b102 Are BuildErrors sufficient here or do we need to make a new NengoException subclass?

This LGTM, but we should make a card to go through the rest of the code base? e..g, There are a bunch of `ValueError` exceptions in `hardware/builder.py`.

arvoelke:
I just noticed that the docstring simulator says the default for `precompute=True`. We should fix that here IMO.

tbekolay:
> We should fix that here IMO.

OK, will do. Running into some issues getting the hardware build to pass so I'll be doing at least one more push anyhow.

PR #204, opened 2019-03-27
Switch to nengo-bones templating

drasmuss:
This replaces #193 as well, since caching and pytest-xdist are now built into the template scripts.


.travis.yml
```diff
@@ -1,38 +1,115 @@
...
 before_install:
-  - source .ci/conda.sh install
+  # export travis_terminate for use in scripts
+  - export -f travis_terminate
+  - export -f _travis_terminate_linux
+  - export -f _travis_terminate_osx
+  - export -f _travis_terminate_unix
+  - export -f _travis_terminate_windows
+  # install/run nengo-bones
+  - pip install git+https://github.com/nengo/nengo-bones.git@docs
```
arvoelke: NTS: This is what's currently causing the build to fail, because `docs` is not rebased onto `template-ci` (but `docs-cleanup` is).

arvoelke: `docs` is now rebased onto `template-ci` and `docs-cleanup` deleted

Issue #217, opened 2019-04-24
Tests running in parallel can still run into concurrency issues


tbekolay:
In 70ecb8a475985846c81bc5e8059056042f855a34 we started running hardware tests in parallel, and made a quick patch to ensure the C file for the snip could be access by two threads at once. It seems that this can [still fail](https://travis-ci.com/nengo/nengo-loihi/jobs/195327799#L1723) so we should try to reproduce that issue and fix it.

arvoelke:
Small update for posterity: to work-around this temporarily, we partially reverted the above commit and switched to `-n 1` py-test workers in #202. But we're still seeing a different related issue (`os.chdir` not changing the directory back in the hardware interface's `__exit__`, and then `coverage.xml` gets put into the wrong place) if `-n` is ommitted altogether.

Still investigating why the directory is not being switched back. This happens consistently [on the same test](https://travis-ci.com/nengo/nengo-loihi/jobs/195940847): `nengo_loihi/hardware/tests/test_allocators.py::test_block_size`.

arvoelke:
The above problem is solved in #218. The larger issue described in e9cd851b01b20b774d58c82a2ba775b531cf2bab is still at hand.

drasmuss:
This seems to have been working OK since https://github.com/nengo/nengo-loihi/commit/ee0ca71fc341685865d26404cc5745bda070e89e#diff-bb7e722a615e3b3171b86e0b952a51fc (hard to say for sure with concurrency issues, but it's at least a pretty rare issue which can be resolved by re-running the build if it ever comes up again).

PR #218, opened 2019-04-26
Fix travis CI

arvoelke:
The first commit makes our `pytest.raises` checks more strict. This is something I was using to debug, and isn't strictly needed by this PR, but IMO is a good thing to do.

The second commit localizes the `os.chdir` to be within a `try-finally` block. This avoids a really weird issue we were seeing as a part of #217 (e.g., [see this job](https://travis-ci.com/nengo/nengo-loihi/jobs/195940847) and look for `nengo_loihi/hardware/tests/test_allocators.py::test_block_size`).

Essentially... if an error happens when the model is being built, then the current working directory (cwd) is permanently changed. More specifically, an error occurs within `HardwareInterface.__init__`, and so `__exit__` is never invoked. As a result, the directory is never changed back. This only occurs for `--target loihi`, and causes `coverage.xml` to be written to the wrong directory. Part of what made this issue weird is that `coverage.xml` is only written to the wrong directory if `-n` is omitted, but not for `-n 1`. This may be a subtle aspect of how pytest spins up its workers.

The second commit also adds a check to pytest's setup/teardown to make sure the directory is not changing. This test would fail on `test_block_size` with `--target loihi` if it were not for the fix.

Note this does *not* solve #217 (i.e., it does not make `nengo-loihi` process safe). The issue explained in the commit message of e9cd851b01b20b774d58c82a2ba775b531cf2bab is still present. Solving this is outside the scope of PR.

### DIFF
--- a/.ci/hardware.sh.template
+++ b/.ci/hardware.sh.template
@@ -20,7 +20,8 @@
         pip install nengo-dl jupyter
         pip install -e .[tests]
         pip install $NENGO_VERSION --upgrade
-        pip install ~/travis-ci/nxsdk-0.8.0.tar.gz
+        cp /nfs/ncl/releases/$NXSDK_VERSION/nxsdk-$NXSDK_VERSION.tar.gz .
+        pip install nxsdk-$NXSDK_VERSION.tar.gz
         SLURM=1 pytest --target loihi --no-hang -v --durations 50 --color=yes -n 1 --cov=nengo_loihi --cov-report=xml --cov-report=term-missing || HW_STATUS=1
         exit \$HW_STATUS
 EOF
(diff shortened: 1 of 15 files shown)


hunse:
I think we should be doing the `chdir` in `create_io_snip`, which gets called when `run` is called, not when the `HardwareInterface` is created. It has the added advantage of keeping all the snip stuff in one place. I don't think there's any need to change the directory in `__init__`.

arvoelke:
Wait why do we even `chdir` at all? It must have been necessary at some point, but that no longer seems to be the case. Everything passes with `nxsdk-0.8.0` without the `chdir`. I added a couple commits to demonstrate.

hunse:
Yeah, I thought I remembered Intel trying to fix that, though I don't ever remember them conclusively saying they did. But if it works without it for 0.8, then let's just get rid of it. We can probably drop support for pre-0.8 by this point. It was definitely necessary with some earlier version.

PR #215, opened 2019-04-17
Update for new nengo-bones release

drasmuss:
**Motivation and context:**

https://github.com/nengo/nengo-bones/pull/14 removes conda from the TravisCI setup, so everything is done with pip instead.  This simplifies and speeds up the build process.  This PR updates `.nengobones.yml` to be consistent with that new approach.

**Interactions with other PRs:**

Depends on https://github.com/nengo/nengo-bones/pull/14

**How has this been tested?**

All TravisCI checks are passing.

**How long should this take to review?**

- Quick (less than 40 lines changed or changes are straightforward)

**Types of changes:**

- Non-code change (touches things like tests, documentation, build scripts)

**Checklist:**

- [x] I have read the **CONTRIBUTING.rst** document.
- [x] I have updated the documentation accordingly.
- [ ] I have included a changelog entry.
- [x] I have added tests to cover my changes.
- [x] I have run the test suite locally and all tests passed.

PR #219, opened 2019-05-07
Template more files with nengo-bones

drasmuss:
Applies the new changes from https://github.com/nengo/nengo-bones/pull/17 to this repo.  This shouldn't result in any functional changes, it is just refactoring/adding files.

Note: based on https://github.com/nengo/nengo-loihi/pull/215


.nengobones.yml
```diff
@@ -1,9 +1,111 @@
...
+    - nengo_loihi *.c
+    - nengo_loihi *.h
+    - nengo_loihi *.template
+  include:
+    - pytest.ini
+  exclude:
+    - .ci/travis_rsa.enc
+
+setup_py:
+  python_requires: ">=3.4"
```
arvoelke: Should this be pegged to 3.5.2?

.nengobones.yml
```diff
@@ -1,9 +1,111 @@
...
+    - scipy
+  classifiers:
+    - "Development Status :: 3 - Alpha"
+    - "Framework :: Nengo"
+    - "Intended Audience :: Science/Research"
+    - "License :: Free for non-commercial use"
+    - "Operating System :: OS Independent"
+    - "Programming Language :: Python :: 3.4"
+    - "Programming Language :: Python :: 3.5"
+    - "Programming Language :: Python :: 3.6"
```
arvoelke: Likewise should the classifiers here be limited to Python 3.5?

.nengobones.yml
drasmuss: I think we want to allow people to use nengo-loihi who are only using it for the emulator, in which case there's no need to pin it to 3.5.2.  We should probably add some TravisCI jobs to cover other versions though, with that in mind.

PR #221, opened 2019-05-23
Fix codespell and update to next bones release

tbekolay:


nengo_loihi/tests/test_validate.py
```diff
@@ -16,7 +16,7 @@ def test_validate_block():
     # too many input axons
     block = LoihiBlock(410)
     block.add_synapse(Synapse(5000))
-    with pytest.raises(BuildError, match="[Ii]nput axon"):
```
hunse: Did codespell start complaining about these? That's kind of annoying, because I did this intentionally in case we change the warning slightly (e.g. to say something like "The input axons" instead of "Input axons"). It's not a big deal, because we can just change the tests if the warning changes, but I am annoyed that codespell complains about this, so I just needed somewhere to complain.

nengo_loihi/builder/connection.py
```diff
@@ -369,10 +369,10 @@ def build_solver(model, solver, conn, rng, sampled_transform):
 
 @Builder.register(NoSolver)
 def build_no_solver(model, solver, conn, rng, sampled_transform):
+    args = (model, solver, conn, rng)
     if nengo_transforms is None:
-        return _build_no_solver(model, solver, conn, rng, sampled_transform)
-    else:
-        return _build_no_solver(model, solver, conn, rng)
+        args += (sampled_transform,)
```
hunse: How does this work with tuples?? I thought they were immutable. I guess here `y += x` is just shorthand for `y = y + x`, which is very strange to me. (If you do `a = (1,2,3); b = a; a += (4,); print(b)`, it prints "(1, 2, 3)", which is different than the `[1, 2, 3, 4]` you get if `a` is a list.)

Anyway, I guess it's fine, I just had never seen that before.

nengo_loihi/tests/test_validate.py
tbekolay: I think it was just `utput` that it complained about, figured I might as well make it consistent for all of them. It would be nice if `pytest.raises` could be case insensitive too, but doesn't seem like that's possible right now

nengo_loihi/builder/connection.py
tbekolay: They are immutable yeah. Under the hood, `+=` results in a call to the left hand side's `__iadd__` method, which can do whatever it wants. I guess for lists, it does the append in-place and for tuples it makes a new tuple and returns the new instance.

PR #223, opened 2019-06-04
Fix test_simulator.test_dt

hunse:
- The x**2 function was hard for these neurons to represent on a
  short time scale, and wasn't necessary for testing `dt`.
  So I've just replaced it with a simple negative function.
- The filters on the reference outputs needed to be chosen better,
  to ensure the same amount of delay in the signal.


tbekolay (approved):
LGTM! Will merge when CI passes.

PR #226, opened 2019-06-07
Multi-chip population spikes

hunse:
Allow population spikes between chips. This allows convolutional connections from chip to chip.

### DIFF
--- a/nengo_loihi/block.py
+++ b/nengo_loihi/block.py
@@ -610,14 +610,21 @@ class Synapse:
         self.indices = indices
 
     def set_full_weights(self, weights):
-        self._set_weights_indices(weights)
+        weights = np.array(weights, copy=False, dtype=np.float32)
+        assert weights.ndim in (1, 2)
+
+        if weights.ndim == 1:
+            indices = np.arange(weights.size)
+        else:
+            indices = None
+
+        self._set_weights_indices(weights, indices=indices)
         assert len(self.weights) == self.n_axons, (
             "Full weights must have different weights for each axon")
 
-        bits = self.idx_bits()
         self.format(
             compression=d(b'Mw==', int),
-            idx_bits=bits,
+            idx_bits=self.idx_bits(),
             fanout_type=d(b'MQ==', int),
             n_synapses=d(b'NjM=', int),
             weight_bits=d(b'Nw==', int),
--- a/nengo_loihi/block.py
+++ b/nengo_loihi/block.py
@@ -354,9 +354,20 @@ class Axon:
         return [self.Spike(axon_id, atom=atom) if axon_id >= 0 else None
                 for axon_id, atom in zip(axon_ids, atoms)]
 
-    def set_axon_map(self, compartment_map, compartment_atoms=None):
-        self.compartment_map = compartment_map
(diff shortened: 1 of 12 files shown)


hunse:
We can use probes with population axons with this change, so we should enable and test that: https://github.com/nengo/nengo-loihi/blob/92d85d3626b37e3b455cae68ea85e4376fc4dc16/nengo_loihi/hardware/builder.py#L438

hunse:
I noticed that in `test_conv_connection`, the ensemble is off-chip, when it should be on-chip. However, there are a few other minor improvements needed for this, namely #232 and allowing neuron slices (which is done in the DVS stuff somewhere).

nengo_loihi/builder/connection.py
```diff
@@ -740,7 +775,8 @@ def build_conv2d_connection(model, conn):
 
     gain = model.params[conn.post_obj.ensemble].gain
     if not np.all(gain == gain[0]):
-        # TODO: support this?
+        # Cannot fold gains into weights, result would not be convolutional.
+        # Therefore, Loihi does not support this if we want to share weights.
```
drasmuss: Couldn't you use something like the decode interneurons to support this? I.e. the convolutional connection would target one set of compartments (with no gain), and then you'd have a second layer implementing the gains.

nengo_loihi/hardware/builder.py
```diff
@@ -473,111 +472,54 @@ def collect_axons(nxsdk_core, core, block, axon, compartment_ids):
     compartment_idxs = np.arange(len(compartment_ids))
     spikes = axon.map_spikes(compartment_idxs)
 
-    all_axons = []  # (compartment, atom, type, tchip_id, tcore_id, taxon_id)
     for compartment_id, spike in zip(compartment_ids, spikes):
+        if spike is None:
+            continue  # means this cx does not route through these axons
```
drasmuss: cx -> compartment

nengo_loihi/block.py
```diff
@@ -598,6 +609,23 @@ def _set_weights_indices(self, weights, indices=None):
         assert len(weights) == len(indices)
         self.indices = indices
 
+    def set_weight_matrix(self, weights):
```
drasmuss: Should combine this and `set_full_weights` into one function.

nengo_loihi/builder/connection.py
hunse: We could definitely use that idea. But our decode-neuron system right now is designed for representing vectors in state-space, so there's multiple neurons per dimension. It'd take some modification.

Even then, there are considerable downsides to this. It changes the dynamics of the network, you need to figure out good spike rates that allow fast information transfer but don't affect the computation due to Loihi aliasing, and it doubles the number of neurons you need. It would be surprising to a user if all these changes happen just because they switch the gains from a homogeneous value to heterogeneous ones. It's also not needed for convnets, which are the main motivation for the Convolution transform.

Maybe at some point we could add an opt-in way to do this, but frankly I'm not sure if it'll ever happen, unless some pressing use-case develops that I'm not forseeing.

drasmuss: Yeah that makes sense to me :+1:

PR #228, opened 2019-06-10
Obfuscate nxsdk information

drasmuss:
Intel has requested that we obfuscate information related to the `nxsdk` API (since it is not publicly available).  This is an attempt at doing so.

Note that the goal here is not to make the obfuscated information unrecoverable, but simply to obscure it from a surface viewing.

There are three major things I was considering in the obfuscation:
1. Hiding any direct calls to `nxsdk` objects
2. Renaming any `nengo-loihi` classes/functions/variables that directly correspond to equivalent `nxsdk` objects
3. Hiding any constant values (e.g. register sizes) that are based on information inside `nxsdk`


nengo_loihi/builder/connection.py
```diff
@@ -404,10 +404,10 @@ def build_chip_connection(model, conn):  # noqa: C901
     # Create random number generator
     rng = np.random.RandomState(model.seeds[conn])
 
-    pre_cx = model.objs[conn.pre_obj]['out']
-    post_cx = model.objs[conn.post_obj]['in']
-    assert isinstance(pre_cx, (LoihiBlock, LoihiInput))
-    assert isinstance(post_cx, (LoihiBlock, Probe))
+    pre_compartment = model.objs[conn.pre_obj]['out']
```
hunse: I think "compartment" here is not accurate, since these are mostly `LoihiBlock` objects, not compartments. I'd be okay with `pre_block` etc., though as you can see in the following line, this can actually be `LoihiBlock` or `LoihiInput`.

nengo_loihi/builder/connection.py
```diff
@@ -724,23 +724,23 @@ def build_conv2d_connection(model, conn):
     pop_type = 32  # TODO: pick this
     new_transform = copy.copy(conn.transform)
     type(new_transform).init.data[new_transform] = weights
-    weights, indices, axon_to_weight_map, cx_bases = conv2d_loihi_weights(
+    weights, indices, axon_to_weight_map, bases = conv2d_loihi_weights(
```
hunse: Maybe "offsets" would be better than "bases", since they're essentially offsets for the weight indices.

nengo_loihi/conv.py
```diff
@@ -114,13 +114,13 @@ def conv2d_loihi_weights(transform):
 
     weights = []
     indices = []
-    cx_bases = np.zeros(input_rows * input_cols, dtype=int)
+    bases = np.zeros(input_rows * input_cols, dtype=int)
```
hunse: Again, I think "offsets" would be better.

nengo_loihi/block.py
```diff
@@ -124,32 +127,32 @@ class Compartment:
...
 
         # parameters common to core
         self.vmin = 0
         self.vmax = np.inf
-        self.noiseMantOffset0 = 0
-        self.noiseExp0 = 0
-        self.noiseAtDendOrVm = 0
+        self.noise_offset = 0
+        self.noise_exp = 0
+        self.noise_dend = 0
```
hunse: I'm pretty sure that a 0 here indicates "dendrites", and a 1 indicates "membrane". So maybe call this "noise_at_membrane" or something like that? Also, it should be documented above (I guess we just forgot before).

nengo_loihi/builder/connection.py
drasmuss: I basically just replaced "cx" with "compartment" throughout 🤷‍♂

PR #229, opened 2019-06-12
Re-enable parallel hardware tests

drasmuss:
The only change from what we had before is that we now do the rendering of the snip template files in a temporary directory (to ensure we don't have multiple simulators trying to write to the same file).  Ran the CI build a bunch of times (see https://travis-ci.com/nengo/nengo-loihi/builds) and didn't see any problems.  Hard to say for sure with these race conditions, I'll keep running it some more, but seems good 🤷‍♂ .

### DIFF
--- a/.nengobones.yml
+++ b/.nengobones.yml
@@ -222,6 +222,10 @@ setup_cfg:
         no ensembles onchip
       test_transforms.py:test_convolution*:
         no ensembles onchip
+      test_synapses.py:test_combined_delay:
+        no ensembles onchip
+      test_synapses.py:test_direct:
+        no ensembles onchip
 
       # accuracy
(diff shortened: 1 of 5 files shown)


drasmuss:
Made a few fixups, and this hasn't failed in 20+ builds now.  I added an extra commit that also copies the include directory (https://github.com/nengo/nengo-loihi/pull/229/commits/bbd4fa56eae42d6d6e21e113881a46720e3aa3cc).  As far as I can tell, this isn't needed (I never saw it fail without this commit).  And I didn't do it before, because we don't know what is in that include directory, so it could be a potentially time-consuming operation to copy it all.  But I added it in to be extra cautious; we could leave it in or take it out 🤷‍♂.

PR #234, opened 2019-07-04
NxSDK 0.8.7

hunse:
Support NxSDK 0.8.7. This is not backwards compatible.


arvoelke:
FWIW I've tried this branch out on `0.9-daily-20190827` and it seems to be working all right!

hunse:
Closed in favour of #255 (which includes this, and also does NxSDK 0.9 support).

PR #236, opened 2019-07-12
Update to support recent changes in nengo core

drasmuss:
Minor changes to get tests passing again after recent updates in nengo master.


nengo_loihi/compat.py
```diff
@@ -47,6 +51,9 @@ def seed_network(*args, **kwargs):
         # point doing anything in this function
         pass
 
+    def make_process_step(process, shape_in, shape_out, dt, rng, dtype=None):
+        return process.make_step(shape_in, shape_out, dt, rng, dtype=dtype)
```
hunse: I think this will only work for synapses, since general processes didn't have a `dtype` argument on `make_step` in 2.8.0. Since we don't currently use this function for anything but synapses, I'm fine to either a) rename this function to `make_synapse_step`, or b) assert that `isinstance(process, Synapse)` here.

tbekolay: I'm going to opt for adding the assert.

PR #240, opened 2019-07-26
Implement Sparse transforms

hunse:
Support `Sparse` transforms for on-chip connections.

Also uses sparse matrices when possible for node->neuron and neuron->neuron connections that have a scalar or diagonal transform.

### DIFF
--- a/.nengobones.yml
+++ b/.nengobones.yml
@@ -35,6 +35,7 @@ setup_py:
   install_req:
     - jinja2
     - nengo>=2.8.0
+    - scipy>=1.2.1
   docs_req:
     - abr_control
     - jupyter
@@ -53,7 +54,6 @@ setup_py:
     - pytest-cov>=2.6.0
     - pytest-xdist>=1.26.0,<1.28.0
     - matplotlib>=2.0
-    - scipy
     - tensorflow<1.14
   classifiers:
     - "Development Status :: 3 - Alpha"
--- a/.nengobones.yml
+++ b/.nengobones.yml
@@ -91,7 +91,6 @@ setup_cfg:
       - superfluous-parens
       - undefined-loop-variable
       - unnecessary-pass
-      - unused-import
       - unused-variable
       - wrong-import-position
     known_third_party:
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -27,12 +27,16 @@ Release history
 - It is now possible to slice the ``pre`` neurons in a neuron->neuron
   connection.
   (`#226 <https://github.com/nengo/nengo-loihi/pull/226>`__)
+- Connections now support ``Sparse`` transforms.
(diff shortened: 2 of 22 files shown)


nengo_loihi/utils/sparse_matrix.py
```diff
@@ -0,0 +1,126 @@
+import warnings
```
drasmuss: What about putting this in `builder/sparse_matrix` rather than `utils/sparse_matrix`? Just feels like more than "utils" to me, since they're required for the connection building.

nengo_loihi/utils/sparse_matrix.py
```diff
@@ -0,0 +1,126 @@
...
+            assert scale.size == matrix.shape[1]
+            diag = scipy_sparse.dia_matrix((scale, 0), shape=(scale.size, scale.size))
+            return matrix.dot(diag)
+        else:
+            return matrix * scale
+    else:
+        return matrix * scale
+
+
+def stack_matrices(matrices, order="v"):
```
drasmuss: An option would be to make `scipy` a dependency of nengo-loihi, then convert all sparse matrices to scipy matrices. That would let us simplify this whole file a lot.  Since nxsdk already requires scipy, this would only make a difference for emulator-only users (and it's not the end of the world for them).

nengo_loihi/utils/sparse_matrix.py
tbekolay: I would also really want to avoid making a whole new submodule (`utils`) for one thing. My dislike of a `utils` submodule in general is also well documented

drasmuss:
Oh, I also rebased this onto https://github.com/nengo/nengo-loihi/pull/241

nengo_loihi/utils/sparse_matrix.py
hunse: Yeah, I'd be fine with that. `nxsdk` doesn't *really* require Scipy I don't think; I'm pretty sure it's just for one of their extra packages. But they way they do their `setup.py` they just put everything they might ever use as "requirements". But that's how everyone is going to be installing. Scipy is also way easier to install than it used to be (on most hardware at least).

nengo_loihi/utils/sparse_matrix.py
hunse: I put them in `utils` because the `is_` functions are used other places than just the `builder/` folder (namely `block.py`). But I'd be okay with just having them in `builder/` anyway (though I would call the file `sparse_matrix_utils.py` I think, just to make it clear that that file is not for building sparse matrices), or having the file in the main directory.

arvoelke:
Should [the documentation](https://www.nengo.ai/nengo-loihi/installation.html) be sync'd up to reflect scipy dependency now?

PR #246, opened 2019-08-14
Fix bug with no inputs to a neuron

hunse:
Some arrays have shape `(1, 0)` which gives them a length of 1, but no size.

### DIFF
--- a/nengo_loihi/tests/test_connection.py
+++ b/nengo_loihi/tests/test_connection.py
@@ -752,3 +752,38 @@ def test_input_synapses(Simulator, allclose, plt):
 
     # Only looking at t < 0.4 as there are weird effects at the end
     assert allclose(ref_filt[t < 0.4], sim_filt[t < 0.4], atol=1.5)
+
+
+@pytest.mark.skipif(nengo_transforms is None, reason="Requires new nengo.transforms")
+def test_sparse_transforms_empty_neurons(Simulator):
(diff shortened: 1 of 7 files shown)


nengo_loihi/tests/test_connection.py
```diff
@@ -749,3 +749,21 @@ def test_input_synapses(Simulator, allclose, plt):
 
     # Only looking at t < 0.4 as there are weird effects at the end
     assert allclose(ref_filt[t < 0.4], sim_filt[t < 0.4], atol=1.5)
+
+
+@pytest.mark.skipif(nengo_transforms is None, reason="Requires new nengo.transforms")
+def test_sparse_transforms(Simulator):
```
hunse: Maybe `test_sparse_transform_empty_neuron` or something like that, since we're specifically testing sparse transforms that have some neurons that get no inputs, not sparse transforms generally. Also, we should put this description in a little docstring.

hunse:
Sounds good. Thanks for the test!

hunse:
Ok, I think I've fixed the issue. :crossed_fingers: The solution was to not build synapses that have no weights, which is something that I meant to do at some point anyway. I had to rebase onto #261, so this will have to wait a bit until that's merged first.

hunse:
Also, I've got this as two commits right now, but I'd be fine squashing them (maybe to something like "Fix sparse connections with no inputs to some neurons").

Issue #63, opened 2018-09-12
Input synapses are currently ignored


tcstewar:
This came up while writing tests for #26 .  That PR changes how inputs are handled when precompute=False, so I wanted to test that the signals are being passed and filtered correctly.  It turns out that it looks like the synapse on an input Connection (i.e. one that goes from a Node to an Ensemble) is currently being ignored.  Here's the test:

```python
@pytest.mark.parametrize('precompute', [False, True])
def test_input_interneurons_running(Simulator, allclose, plt, precompute):
    synapse = 0.1
    with nengo.Network() as model:
        stim = nengo.Node(lambda t: 1 if t % 0.5 < 0.25 else 0)
        ens = nengo.Ensemble(n_neurons=1, dimensions=1,
                             encoders=[[1]],
                             intercepts=[0],
                             max_rates=[40])
        c = nengo.Connection(stim, ens, synapse=synapse)   # This is the synapse that is being ignored
        p_stim = nengo.Probe(stim)
        p_neurons = nengo.Probe(ens.neurons, synapse=0.1)
    with Simulator(model, precompute=precompute) as sim:
        sim.run(1.0)
    with nengo.Simulator(model) as ref:
        ref.run(1.0)

    plt.plot(sim.trange(), sim.data[p_neurons], label='nengo_loihi')
    plt.plot(sim.trange(), ref.data[p_neurons], label='nengo')
    plt.legend(loc='best')
    plt.twinx()
    plt.plot(sim.trange(), sim.data[p_stim])
```

Here is the result (for both precompute=True and precompute=False):

![image](https://user-images.githubusercontent.com/2380772/45404245-14e26e00-b62c-11e8-9162-6d0a4e330560.png)

If I modify the test such that reference nengo ignores the input synapse by setting `c.synapse=None`, then we get this

![image](https://user-images.githubusercontent.com/2380772/45404290-4e1ade00-b62c-11e8-9bf4-cffcb58668d1.png)

The synapse is being set on the Connection that's being passed to the builder ( https://github.com/nengo/nengo-loihi/blob/master/nengo_loihi/splitter.py#L181 ), but I'm not sure how that ends up being used.

This affects both the emulator (target='sim') and the chip itself (target='loihi')

[...]

PR #250, opened 2019-10-09
Add test for input synapses

tbekolay:
This adds the test described in #63, with some additional filtering (Nengo Loihi drifts a little bit over time) and with an assertion at the end.

It's based on #247, which should be merged first.

As this just adds a test, I don't believe it needs a changelog entry.

### DIFF
--- a/nengo_loihi/tests/test_connection.py
+++ b/nengo_loihi/tests/test_connection.py
@@ -719,3 +719,33 @@ def test_sparse_ens_ens(Simulator, seed, plt, allclose):
 
     assert allclose(sim.data[bp][:, 0], -0.8 * sim.data[up][:, 2], atol=0.2)
     assert allclose(sim.data[bp][:, 1], 0.6 * sim.data[up][:, 0], atol=0.2)
+
+
+def test_input_synapses(Simulator, allclose, plt):
+    synapse = 0.1
+    with nengo.Network() as net:
+        stim = nengo.Node(lambda t: 1 if t % 0.5 < 0.25 else 0)
+        ens = nengo.Ensemble(
+            n_neurons=1, dimensions=1, encoders=[[1]], intercepts=[0], max_rates=[50]
+        )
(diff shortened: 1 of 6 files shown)


nengo_loihi/tests/test_connection.py
```diff
@@ -719,3 +719,33 @@ def test_sparse_ens_ens(Simulator, seed, plt, allclose):
...
+
+    t = sim.trange()
+    ref_filt = nengo.Alpha(0.03).filtfilt(ref.data[p_neurons])
+    sim_filt = nengo.Alpha(0.03).filtfilt(sim.data[p_neurons])
+
+    plt.plot(t, ref_filt, label="nengo")
+    plt.plot(t, sim_filt, label="nengo_loihi")
+    plt.legend(loc="best")
+    plt.twinx()
+    plt.plot(t, sim.data[p_stim], c="k")
```
arvoelke: Do you have this plot handy to upload here? Could be useful in the future to know what you mean about the weird effects at the end.

nengo_loihi/tests/test_conv.py
```diff
@@ -707,7 +707,9 @@ def test_conv_split(Simulator, rng, plt, allclose):
 
 @pytest.mark.skipif(nengo_transforms is None, reason="Requires new nengo.transforms")
 def test_conv_preslice(Simulator, plt):
-    from nengo._vendor.npconv2d.conv2d import conv2d
+    from nengo._vendor.npconv2d.conv2d import (  # pylint: disable=import-outside-toplevel
```
tbekolay: I would just turn these into a `pytest.importorskip`, then remove the `skipif`

nengo_loihi/tests/test_connection.py
tbekolay: I don't, but the weird effects are basically that Nengo Loihi's decoded value drops toward zero in the last 50 or so ms of the simulation, regardless of the length of the simulation. I didn't look into it much detail to know why.

PR #255, opened 2019-10-22
NxSDK 0.9.0

hunse:
Support NxSDK 0.9.0. No changes are actually required, but we now do not warn for and test against 0.9.0.

I also updated `HardwareInterface` to store the min and max NxSDK versions on the class, so these can be checked. We now no longer have to remember to update the test in `hardware/tests/test_interface.py` each time we change the NxSDK version.

Depends on #234.


hunse:
TODO:

- [ ] There are a number of warnings that functions we're using are now deprecated with NxSDK 0.9.0. Go through the test suite warnings and move over to the new functions where appropriate.

PR #259, opened 2019-11-08
Update to support recent versions of various packages

drasmuss:
This brings Nengo Loihi up-to-date with Nengo core, Nengo Bones, Nengo DL, and TensorFlow.

Note that this depends on a PR in Nengo DL (https://github.com/nengo/nengo-dl/pull/108) and Nengo Extras (https://github.com/nengo/nengo-extras/pull/90).

I didn't do the Nengo DL updates in a backwards compatible way. We could probablyyyy make it work if we wanted to, but it'd be pretty messy. And since Nengo Loihi is a pretty rapidly changing project, requiring users to upgrade their Nengo DL installation doesn't seem too bad.


tbekolay:
The coverage changes are due to no longer coverage some compatibility stuff as none of the TravisCI jobs use Nengo 2.8.0 anymore. I made #263 to discuss that so that this PR isn't blocked.

PR #260, opened 2019-11-15
Host snips

hunse:
This is a partial version of #256, that just does the host snips and some more basic speed improvements (the other PR had some more experimental speed improvements, too, which need more testing).

It's been rebased onto #255 and cleaned up.


hunse:
BTW, the build is actually passing, it's just some stupid docs stuff that failed.

nengo_loihi/hardware/snips/nengo_host.cc.template
```diff
@@ -102,7 +110,8 @@ class NengoHostProcess : public ConcurrentHostSnip {
             }
 
             // write packets to chip (PACKET_SIZE == sizeof(int32_t))
-            writeChannel(input_channel, buffer, n_read / sizeof(int32_t));
+            size_t n_packets = ceil_div<size_t>(n_read, 4*PACKET_SIZE);
```
kinjalpatel27: The packet size in bytes `4*PACKET_SIZE` can be converted to `PACKET_BYTE_SIZE` to have better clarity on why the packet size is multiplied by 4

nengo_loihi/simulator.py
```diff
@@ -509,13 +511,18 @@ def loihi_precomputed_host_only(steps):
...
 
                 def loihi_bidirectional_with_host(steps):
                     loihi.run_steps(steps, blocking=False)
+                    time0 = timeit.default_timer()
                     for _ in range(steps):
                         host.step()
                         self._host2chip(loihi)
                         self._chip2host(loihi)
+                    time1 = timeit.default_timer()
+                    self.timers["snips"] += time1 - time0
```
kinjalpatel27: The variable names `time0` and `time1` can be renamed to something more descriptive e.g. `time_snips_start` and `time_snips_end`. Also, similar to `self.timers["steps"]`, `time1` can be eliminated by `self.timers["snips"] += timeit.default_timer() - time0` unless there is other plans to use `time1`

nengo_loihi/simulator.py
```diff
@@ -563,6 +575,7 @@ def run_steps(self, steps):
                 )
             raise
 
+        self.timers["steps"] += timeit.default_timer() - time0
```
kinjalpatel27: The variable names `time0` can be renamed to something more descriptive e.g. `time_steps_start`

nengo_loihi/hardware/snips/nengo_host.cc.template
```diff
@@ -0,0 +1,147 @@
...
+    int32_t buffer[BUFFER_SIZE];
+    int server_socket;
+    int client_socket;
+
+  public:
+    NengoHostProcess() {
+        // --- set up socket to communicate with superhost
+        // This machine will act as the server, and the superhost will
+        // connect in as the client
+        server_socket = socket(AF_INET, SOCK_STREAM, 0);
```
kinjalpatel27: For socket programming, I found that it is common to check for errors from functions execution. It might not be needed here probably(not sure). But just in case, the usual checks include if the output of the function is less than 0. e.g.
`if (server_socket < 0)
     std::cout << "Failed to open the socket" << std::endl;
`
The error output is `-1` for `socket`,`bind`,`listen` ,`accept`

nengo_loihi/hardware/interface.py
```diff
@@ -153,6 +166,23 @@ def run_steps(self, steps, blocking=True):
         # start the board running the desired number of steps
         d_get(self.nxsdk_board, b"cnVu")(steps, **{d(b"YVN5bmM="): not blocking})
 
+        # connect to host socket
+        if self.host_socket is not None and not self.host_socket_connected:
+            # pause to allow host snip to start and listen for connection
+            time.sleep(0.1)
+
+            host_address = self.nxsdk_board.executor._host_coordinator.hostAddr
```
kinjalpatel27: I think we discussed this to check if the `nxsdk` attributes need to be obfuscated or not

nengo_loihi/simulator.py
```diff
@@ -365,26 +365,34 @@ def _collect_receiver_info(self):
...
+
+                if spike_target is not None:
+                    spike_idxs = x.nonzero()[0]
+                    spikes.append((spike_target, ti, spike_idxs))
+
+                if error_synapse is not None:
+                    errors_ti = errors.get(ti, None)
+                    if errors_ti is None:
+                        errors_ti = OrderedDict()
+                        errors[ti] = errors_ti
```
kinjalpatel27: As before, line 387 and 388 can be combined as `errors_ti = errors.setdefault(ti, OrderedDict())`.

nengo_loihi/hardware/snips/nengo_host.cc.template
```diff
@@ -70,48 +72,44 @@ class NengoHostProcess : public ConcurrentHostSnip {
...
-            if (n_read % sizeof(int32_t) != 0) {
-                std::cout << "Did not read full int32s from socket" << std::endl;
-                break;
-            }
+            // read packets from superhost socket
+            // first read is blocking, so we wait for packets
+#if DEBUG
+            std::cout << "Host: Preparing to read socket" << std::endl;
+#endif
+            ssize_t n_read = recv(client_socket, buffer, 4*PACKET_SIZE, 0);
```
kinjalpatel27: The packet size in bytes `4*PACKET_SIZE` can be converted to `PACKET_BYTE_SIZE` to have better clarity on why the packet size is multiplied by 4

nengo_loihi/hardware/snips/nengo_host.cc.template
```diff
@@ -70,48 +72,44 @@ class NengoHostProcess : public ConcurrentHostSnip {
...
+                }
 
-            // write packets to chip (PACKET_SIZE == sizeof(int32_t))
-            size_t n_packets = ceil_div<size_t>(n_read, 4*PACKET_SIZE);
-            writeChannel(input_channel, buffer, n_packets);
+                // write packet to chip
+                writeChannel(input_channel, buffer, 1);
+
+                // subsequent reads are non-blocking, so we just read what's there
+                n_read = recv(client_socket, buffer, 4*PACKET_SIZE, MSG_DONTWAIT);
```
kinjalpatel27: The packet size in bytes `4*PACKET_SIZE` can be converted to `PACKET_BYTE_SIZE` to have better clarity on why the packet size is multiplied by 4

nengo_loihi/hardware/interface.py
```diff
@@ -215,13 +215,26 @@ def _chip2host_monitor(self, probes_receivers):
...
+        n_waits = 0  # number of times we've had to wait for more data
+
         recv_size = 4096  # python docs recommend small power of 2, e.g. 4096
         received = self.host_socket.recv(recv_size)  # blocking recv call
         data = received
-        while len(received) == recv_size:
-            received = self.host_socket.recv(recv_size, flags=socket.MSG_DONTWAIT)
-            data += received
-        assert len(data) == 4 * self.nengo_io_c2h_count
+        while len(data) < expected_bytes and n_waits < 10:
```
kinjalpatel27: It may be helpful in the future to add a comment on why `n_waits < 10`, even if it is chosen at random

nengo_loihi/hardware/interface.py
```diff
@@ -215,13 +215,26 @@ def _chip2host_monitor(self, probes_receivers):
     def _chip2host_snips(self, probes_receivers):
         assert self.host_socket_connected
 
+        expected_bytes = 4 * self.nengo_io_c2h_count
```
kinjalpatel27: The multiplication of size with 4 can be replaced with multiplication with a constant describing 4 e.g. `bytes_per_int32` or something better.

nengo_loihi/hardware/interface.py
```diff
@@ -582,25 +599,33 @@ def create_io_snip(self):
...
         self.nengo_io_c2h = d_get(self.nxsdk_board, b"Y3JlYXRlQ2hhbm5lbA==")(
-            b"nengo_io_c2h",  # name of the channel
+            b"nengo_io_c2h",  # channel name
             **{
-                d(b"bnVtRWxlbWVudHM="): n_outputs,  # size of the channel (in elements)
-                d(b"bWVzc2FnZVNpemU="): 4,  # size of one packet (in bytes)
+                # channel size (in elements)
+                d(b"bnVtRWxlbWVudHM="): n_outputs,
+                # size of one packet (in bytes)
+                d(b"bWVzc2FnZVNpemU="): 4 * self.channel_packet_size,
```
kinjalpatel27: The multiplication of size with 4 can be replaced with multiplication with a constant describing 4 e.g. `bytes_per_int32` or something better. (line 614, 626)

nengo_loihi/simulator.py
hunse: True. In my opinion, though, it's easier/clearer to read as two lines. `setdefault` implies that it could be already set, whereas inside this `if` block we know it's not.

Originally, I was doing this whole `if` block with `setdefault` instead (i.e., replacing lines 385-388 with `errors_ti = errors.setdefault(ti, OrderedDict())`). Then I realized that that makes the `OrderedDict` every time, even if it's not used (i.e. even if `ti in errors`), which is wasteful. So that's why I made the `if` block.

nengo_loihi/hardware/interface.py
hunse: We should probably obfuscate just to be safe and consistent (though the main point of obfuscation is to stop stuff coming up in Google, and I don't think anybody is going to be Googling this).

nengo_loihi/hardware/snips/nengo_host.cc.template
hunse: We could have `PACKET_ELEMENTS` for the size in `int32`s, and `PACKET_BYTES` for the size in bytes.

nengo_loihi/hardware/interface.py
hunse: It might be good to have a class-level variable for this. Each wait is 1 ms, so we could have `snip_max_interpacket_wait_ms` or something.

nengo_loihi/hardware/interface.py
hunse: Maybe where we define `channel_packet_size`, we also have `channel_bytes_per_element`.

Also, if we rename `PACKET_SIZE` to `PACKET_ELEMENTS` in the snips (as I was musing about above), then we should rename `channel_packet_size` to `channel_packet_elements`.

nengo_loihi/hardware/interface.py
hunse: `channel_bytes_per_element`

nengo_loihi/hardware/snips/nengo_host.cc.template
kinjalpatel27: To add the error statements, what should I be doing after the print statement?
Option 1) `return -1`. Although I haven't been able to figure out who would be calling `NengoHostProcess`  and if returning a particular value would mean anything.
Option 2) have a `while`  loop to ensure the success before moving on, but that would still require a block for `timeout` and a `return` statement

hunse: Just do what you suggested first: Print an error message to `cout`, and then return (option 1). This is in the constructor, so I'm not sure if you can return `-1`.

Also, I'm worried that even if you return early from the constructor, the host snip `run` function will still get called. `client_socket` gets set right at the end of the constructor, so if we initialize both sockets to `-1` and have an error at the start of `run` if either one is < 0, then we should be good.

nengo_loihi/splitter.py
```diff
@@ -4,184 +4,280 @@
...
-                    raise BuildError(
-                        "Pre ensemble (%r) of error "
-                        "connection (%r) must not be configured "
-                        "as on_chip." % (pre, conn)
-                    )
-                self._chip_objects.remove(pre)
+        # On-chip convolutional connections are also not supported with
+        # precompute=True because the BasicSpikeGenerator cannot send population
+        # spikes correctly, meaning we have to use SNIPs.
+        # TODO: check with Eric if this logic is correct.
```
hunse: First of all, it's only host->chip connections that are the problem. Connections that are fully on the chip are fine. So if someone sets `precompute=True` on a network that only has convolution connections that are fully on the chip, will this be a problem?

Second, we do add support for BasicSpikeGenerator sending spikes to the chip in #261, so this logic will have to change once we get there.

nengo_loihi/tests/test_conv.py
```diff
@@ -460,7 +460,7 @@ def test_conv_connection(channels, channels_last, Simulator, seed, rng, plt, all
         sim_real.run(pres_time)
     real_out = sim_real.data[bp].mean(axis=0).reshape(output_shape.shape)
 
-    with Simulator(model, dt=dt, precompute=False) as sim_loihi:
```
hunse: Personally, I like being explicit in tests about whether we're testing with `precompute=True` or `precompute=False`. It makes it a lot easier to know what is going on in the test.

nengo_loihi/splitter.py
```diff
@@ -4,184 +4,280 @@
...
+        )
+        self.precompute = precompute
+        if self.precompute is None:
+            self.precompute = self.precomputable()
+
+    def precomputable(self, obj=None):
+        return self._precomputable.precomputable(obj=obj)
+
+    def on_chip(self, obj):
+        return self.hostchip.on_chip(obj)
```
hunse: I think it would be great to have @arvoelke look at these splitter changes, because I think he wrote most of the original code that's being changed.

nengo_loihi/simulator.py
```diff
@@ -226,6 +226,7 @@ def __del__(self):
     def __enter__(self):
         for sim in self.sims.values():
             sim.__enter__()
+        self._runner = StepRunner(self.model, self.sims, self.precompute, self.timers)
```
hunse: Does this mean that the user has to use the simulator in a `with` block, and we no longer support something like:
```
sim = nengo_loihi.Simulator(net)
sim.run(1.0)
sim.close()
```

nengo_loihi/hardware/interface.py
```diff
@@ -427,6 +433,7 @@ def close(self):
 
         if self.nxsdk_board is not None:
             d_func(self.nxsdk_board, b"ZGlzY29ubmVjdA==")
+            self.nxsdk_board = None
```
hunse: Can we note in the commit message that closing the simulator now gets rid of the `nxsdk_board`? This might affect some of my scripts. (My guess is no other users are accessing `nxsdk_board`, so probably doesn't need a changelog entry.)

CHANGES.rst
```diff
@@ -36,9 +36,8 @@ Release history
   testing). We also better detect when we can use ``precompute=True``.
   (`#260 <https://github.com/nengo/nengo-loihi/pull/260>`__)
 - ``Simulator`` now includes a ``.timers`` dictionary that records the wall time
-  taken by particular parts of the simulation. Currently, the main timer available is
-  ``"steps"``, which reports the amount of time taken to run the simulation steps
-  (not including connection time).
+  taken by all simulation steps (``"steps"``) and, for models using SNIPs, the time
+  taken inside SNIPs (``"snips"``).
```
hunse: This isn't quite accurate, and is part of the reason I didn't talk about "snips" here (I made it more for internal use rather than for users). Basically, it measures the total time after the board has been fully initialized/started (i.e. `HardwareInterface.run_steps` is called), in the case that we're using SNIPs. So it's not just "inside SNIPs", it's really all the time on board and chip, just without startup time.

nengo_loihi/hardware/interface.py
```diff
@@ -105,8 +106,9 @@ def __enter__(self):
             raise SimulationError(
                 "Loihi interface has been closed and cannot be reopened."
             )
-        if not self.is_built:
-            self.build()
+
+        if not self.connected:
+            self.connect()
```
hunse: So the reason I avoided this is because I didn't want to force the user to connect to the board when they do `with nengo_loihi.Simulator(...) as sim:`. One example of where you want to avoid this is when doing power measurement with a large model. Since you want to measure power as close to actual run time as possible, my process was something like this:
```
with nengo_loihi.Simulator(big_model) as sim:
    # big model has now built

    # measure idle power (involves connecting to the board, 
    # and I'm not sure if we can have two connections to the board at once)

    # run the big model
    sim.run(...)
```

nengo_loihi/hardware/builder.py
```diff
@@ -379,11 +379,13 @@ def build_input(nxsdk_core, core, spike_input, compartment_idxs):
     nxsdk_board.spike_inputs[spike_input] = loihi_input
 
     # add any pre-existing spikes to spikegen
-    for t in spike_input.spike_times():
+    times = spike_input.spike_times()
+    if len(times) > 0:
```
hunse: Was this assert a big time sink? The only downside to this is that `times` can't be a generator now, but if this is working, then I guess it wasn't a generator anyway, so we can always change accordingly if we want it to be a generator in the future.

nengo_loihi/hardware/interface.py
```diff
@@ -690,6 +633,70 @@ def create_io_snip(self):
         self.nengo_io_snip_range = snip_range
 
 
+class HostSocket:
+    recv_retries = 10
```
hunse: This doesn't seem to be used.

nengo_loihi/hardware/interface.py
```diff
@@ -690,6 +633,70 @@ def create_io_snip(self):
...
+        )
+
+        return np.frombuffer(data, dtype=np.int32)
+
+    def send_all(self, data):
+        msg_bytes = struct.pack("%di" % len(data), *data)
+        msg_len = len(msg_bytes)
+        bytes_sent = 0
+        while bytes_sent < msg_len:
+            _, ready, _ = select([], [self.socket], [])
```
hunse: Should we have a timeout here, maybe 10 s? (we can make it configurable, too)

nengo_loihi/builder/inputs.py
```diff
@@ -43,56 +41,53 @@ def receive(self, t, x):
...
-    def clear(self):
-        self.spikes.clear()
-
-    def receive(self, t, x):
-        assert len(self.spikes) == 0 or t > self.spikes[-1][0]
-        assert x.ndim == 1
-        self.spikes.append((t, x.nonzero()[0]))
-
     def update(self, t):
         raise SimulationError("ChipReceiveNodes should not be run")
```
arvoelke: Minor nitpick: `ChipReceiveNodes` -> `type(self).__name__` since this method is being inherited.

nengo_loihi/decode_neurons.py
```diff
@@ -19,6 +20,7 @@ class DecodeNeurons:
     ----------
     dt : float
         Time step used by the simulator.
+    is_input : bool, optional (Default: False)
```
arvoelke: Should this be here? Doesn't look like the constructor accepts this.

nengo_loihi/decode_neurons.py
```diff
@@ -178,16 +185,27 @@ def get_ensemble(self, dim):
                 "Input neurons with more than one neuron per dimension"
             )
 
+        for param in [
+            "n_neurons",
+            "dimensions",
+            "neuron_type",
+            "encoders",
+            "gain",
+            "bias",
```
arvoelke: Is this checking necessary? As far as I can tell this is only being used internally, and only with the `add_to_container` kwarg? If a developer did specify one of these then it should throw an error further down when the ensemble is created. It's also missing some like `max_rates` and `intercepts`.

nengo_loihi/decode_neurons.py
tbekolay: No, I had this here for a while then moved it to a subclass. Will fix.

nengo_loihi/decode_neurons.py
tbekolay: The intention behind this commit is to make `DecodeNeurons` "user accessible", so users could pass any ensemble arguments they want. The reason to be defensive here is because the error further down will be confusing, it'll be a `TypeError` that'll make it seem like they passed something in twice, but in actuality we pass it in once and they pass it in once, so it just won't work.

We don't pass in everything though, including `max_rates` and `intercepts`, though we should change this in some way because we're now tripping the warning for setting gain, bias, and one of max_rates/intercepts, which we shouldn't.

nengo_loihi/splitter.py
tbekolay: Yes, the chip->chip connections will fail with this. The reason I thought that that would be OK is because all of the tests that we were doing using convolutions explicitly set `precompute=False`, so presumably anything with convolutions isn't precomputable. But if that's not the case then the comment is incorrect and we should only be checking for host->chip convolutional connections, which is easily changed.

Should be trivial to make changes to this logic when things change, don't think this addition makes a difference.

nengo_loihi/hardware/interface.py
```diff
@@ -56,31 +67,70 @@ def __init__(
...
 
-        self.build(model, allocator=allocator, seed=seed)
+        # If we're using snips, tag all probes as being snip-based,
+        # as having normal probes at the same time as snips causes problems.
+        # This must be done before the build process to ensure information
+        # is stored properly.
+        if self.use_snips:
+            for block in self.model.blocks:
+                for probe in block.probes:
+                    probe.use_snip = True
```
arvoelke: Does this mutate the underlying model? Would this cause an issue if the same model was run multiple consecutive times in separate simulators with separate snip settings?

nengo_loihi/tests/test_conv.py
tbekolay: If there were a comment here explaining why it has to be precomputed for test purposes I could understand that, but my understanding/assumption is that it doesn't work with `precompute=True`, but we were defaulting to `precompute=True`, so that seems like an oversight in the splitter rather than something we did intentionally. If we're trying to test something with/without precomputing I would expect this test to be parametrized and tested with both precomputing and not.

nengo_loihi/hardware/interface.py
arvoelke: Oh nevermind this is a nengo-loihi probe from `block.py`. Not a core nengo probe.

nengo_loihi/simulator.py
tbekolay: That's true, this is something I hadn't considered. This is probably a conversation we should have team-wide, as right now the assumption is that constructing the simulator also "opens" it, `__enter__` does nothing, and `__exit__` calls `close`. If we want a two-step initialization, like it sounds like has some advantages in Nengo Loihi, then we need to add an `open` method that people need to call, which we should also add to all backends.

nengo_loihi/hardware/interface.py
tbekolay: Sure, can do.

nengo_loihi/hardware/interface.py
arvoelke: Not sure if this is relevant, but I was finding with the newer NxSDK's I could remove the `self.connect()` call and it would still automatically / lazily connect later on.

CHANGES.rst
tbekolay: I think the key name `"snips"` is probably not the right one then. We'll discuss offline.

nengo_loihi/hardware/interface.py
tbekolay: So right now you can do this by making the sim first and opening the context manager later. As you pointed out in another comment, probably we should make it easier to do this without a context manager by adding an `open` method as well.

nengo_loihi/hardware/builder.py
tbekolay: Since you made changes here for speed reasons I assumed it was a critical loop, so any savings would be worth it. It would probably be good for it to be a generator, and there are other ways to figure out if there are any spikes without doing `len`, so I'll look at this in more detail.

nengo_loihi/hardware/interface.py
tbekolay: Ah yeah, I added in the retry counting but forgot to add the condition to the while loop. Good catch!

nengo_loihi/hardware/interface.py
tbekolay: I don't remember where I read this, but I thought sends always work and can't time out because they just fill a buffer, or something like that... in searching around for a reference on that, I see that the native `socket` library has a [`sendall`](https://docs.python.org/3/library/socket.html#socket.socket.sendall) function that we should likely use instead, as it's implemented in C and likely faster / more reliable.

nengo_loihi/builder/inputs.py
tbekolay: Good call! Will change.

nengo_loihi/splitter.py
```diff
@@ -4,184 +4,280 @@
...
+    """Creates a set of directives to guide the builder.
+
+    Parameters
+    ----------
+    network : Network
+        The original Network supplied to the builder.
+    precompute : bool, optional (Default: None)
+        Whether model inputs should be precomputed to speed up simulation.
+        The splitter will determine precomputable objects regardless,
+        but if
```
arvoelke: Docstring stops abruptly. Attributes also empty.

nengo_loihi/passthrough.py
```diff
@@ -296,15 +292,18 @@ def _find_clusters(self):
...
         """Split a Cluster."""
         assert cluster not in self._already_split
         self._already_split.add(cluster)
 
-        onchip_input = any(base_obj(c.pre) not in self.ignore for c in cluster.conns_in)
-        onchip_output = any(
-            base_obj(c.post) not in self.ignore for c in cluster.conns_out
-        )
+        onchip_input = any(self._on_chip(base_obj(c.pre)) for c in cluster.conns_in)
+        onchip_output = any(self._on_chip(base_obj(c.post)) for c in cluster.conns_out)
```
arvoelke: Note to self: check that this is functionally equivalent.

nengo_loihi/simulator.py
tbekolay: Decision is to move the `StepRunner` creation to `__init__` and make a card for adding `open` methods for all backends for later.

CHANGES.rst
tbekolay: Decision made in the meeting is to merge these two timers so that we always do our best to only profile the run time (which in most cases is with what we were calling `"steps"` but when using snips is what we were calling `"snips"`). Then, we will also add a timer to measure build time (which is minimal effort) and one to measure startup time. The startup time should capture the difference between `steps` and `snips` that is currently in this PR.

nengo_loihi/decode_neurons.py
tbekolay: Decision from the meeting is to remove the `**kwargs` here and just allow passing in `add_to_container`. This makes things simpler for us, and is if anything more convenient for the user, as they can just modify parameters once they have the ensemble if they like.

nengo_loihi/tests/test_conv.py
tbekolay: We agreed to leave this without specifying `precompute`.

nengo_loihi/simulator.py
tbekolay: Done in 96bd3779

nengo_loihi/hardware/interface.py
tbekolay: Done in 51262580 (until it gets squashed in)

nengo_loihi/hardware/interface.py
tbekolay: Deferring entering the context manager is the way to do this now:

```
sim = nengo_loihi.Simulator(big_model)
# big model has now built

# measure idle power (involves connecting to the board, 
# and I'm not sure if we can have two connections to the board at once)

with nengo_loihi.Simulator(big_model) as sim:
    # run the big model
    sim.run(...)
```

nengo_loihi/hardware/interface.py
tbekolay: Done in 22bd6f67

nengo_loihi/hardware/interface.py
tbekolay: Done in a518db1f

nengo_loihi/builder/inputs.py
tbekolay: Done in 42cd76eb

nengo_loihi/decode_neurons.py
tbekolay: Done in fe055cee

nengo_loihi/decode_neurons.py
tbekolay: Done in fe055cee

nengo_loihi/splitter.py
tbekolay: Fixed in 3ffed333

nengo_loihi/splitter.py
tbekolay: Done in e77ece72

nengo_loihi/hardware/builder.py
tbekolay: Resolved in dd4afcea.

nengo_loihi/inputs.py
```diff
@@ -14,17 +16,25 @@ class SpikeInput(LoihiInput):
...
+        assert (
+            ix == len(self._ti) or self._ti[ix] != ti
+        ), "spikes for time %d have already been added" % (ti,)
+        self._ti.insert(ix, ti)
+        self._spikes[ti] = spike_idxs
+
+    @property
     def spike_times(self):
-        return sorted(self.spikes)
+        return self._ti
```
hunse: So the idea here is to keep the list of spike times sorted as we go, rather than sorting when requested? Why is this better? If we were requesting spike times a lot, I could see the rationale, but we only do it once per build (in the `build_input` function).

hunse:
I looked through the new commits. Just the one question, otherwise they all look good.

nengo_loihi/inputs.py
tbekolay: Hm... I think because this PR was about speeding things up I assumed that this part of the diff was also about speeding things up, but looking back this was touched for other reasons. I'll revert.

tbekolay: Reverted in 16f303151d107de9d8cdf3980160cd8bd7969a64.

CHANGES.rst
tbekolay: Done in 08dbb8f16d2477eb61f49a3e4e9bec01fb872e30

nengo_loihi/splitter.py
```diff
@@ -4,184 +4,280 @@
...
                 pre = base_obj(conn.pre)
-                if self.on_chip(pre):
-                    raise BuildError(
-                        "Cannot precompute input, " "as it is dependent on output"
-                    )
-                mark_precomputable(pre)
+                if not self.hostchip.on_chip(pre):
+                    self.mark_precomputable(pre, queue)
+                else:
+                    # We found a closure! The model is not precomputable.
```
arvoelke: Very minor: I wouldn't say that 'we found a closure'. In fact `pre` could just be some dangling ensemble on the chip that isn't receiving any input and isn't connected to anything else. We're doing the right thing in this case by detecting that it's not precomputable, since input to the chip (the first values in the queue) depend on output from the chip (the above `conn`). Think it's more accurate to say 'We found an input to the chip that relies on an output from the chip. The host model is not precomputable.'

nengo_loihi/splitter.py
```diff
@@ -4,184 +4,280 @@
...
-        conns = (
-            set(self.network.all_connections) | self.passthrough.to_add
-        ) - self.passthrough.to_remove
+        # Forwards and backwards adjacency lists
+        pre_to_conn = defaultdict(list)
+        post_to_conn = defaultdict(list)
 
         # Initialize queue with the pre objects on host->chip connections.
         # We assume that all `conn.pre` objects are pre-computable, and then
         # raise an error later if one of them turns out to rely on chip output.
```
arvoelke: Nit: we no longer raise an error. Suggested: "raise an error" -> "set a flag to mark the host model as not precomputable".

nengo_loihi/splitter.py
arvoelke: Done :)

nengo_loihi/splitter.py
tbekolay: That makes sense, I changed it. Well, I say it makes sense, but it makes sense only after thinking real hard about it :dizzy_face:

nengo_loihi/splitter.py
tbekolay: Right you are, thanks!

PR #261, opened 2019-11-15
Dummy axons work properly on chip

hunse:
We were having problems with dummy axons still doing something on the chip, even though they shouldn't. This PR fixes that. I added the test in a separate commit first, so that it's easy to confirm the test fails before the fix and passes after it.

The latter two commits improve support for population axons. The first one allows users to make a connection use pop16 axons instead of the default pop32. This uses fewer axon slots, so that connections can have more axons, but adds some constraints to how these axons are used. (In convolutional networks, this functionally amounts to having to use numbers of filters that are a multiple of 4, with `channels_last=True`. This allows the axon compartment offset ("base") to be a multiple of 4, which is the requirement for pop16. We could make copies of the weights in the case that we need offsets that are not multiples of 4, such that we have up to 4 copies each for a different modulo of the offset. See #262.)

The last commit allows population axons to be used on input connections when `precompute=True` (since NxSDK's spike generators now support them).

Based on #260.

### DIFF
--- a/nengo_loihi/config.py
+++ b/nengo_loihi/config.py
@@ -11,6 +11,10 @@ def add_params(network):
       * ``on_chip``: Whether the ensemble should be simulated
         on a Loihi chip. Marking specific ensembles for simulation
         off of a Loihi chip can help with debugging.
+    `nengo.Connection`
+      * ``pop_type``: The axon format when using population spikes, which are only
+        used for convolutional connections. Must be an int between 16 and 32.
+        By default, we use ``pop_type`` 32.
 
     Examples
     --------
@@ -28,6 +32,10 @@ def add_params(network):
     ens_cfg = config[nengo.Ensemble]
     if "on_chip" not in ens_cfg._extra_params:
         ens_cfg.set_param("on_chip", Parameter("on_chip", default=None, optional=True))
+
+    conn_cfg = config[nengo.Connection]
+    if "pop_type" not in conn_cfg._extra_params:
+        conn_cfg.set_param("pop_type", Parameter("pop_type", default=32, optional=True))
 
 
 def set_defaults():
--- a/nengo_loihi/hardware/nxsdk_objects.py
+++ b/nengo_loihi/hardware/nxsdk_objects.py
@@ -245,6 +245,8 @@ class LoihiSpikeInput:
     atom : np.int32
         The population index (atom), used if this axon sends population spikes
         (i.e. axon_type != 0).
+    atom_bits_extra : np.int32
+        The number of extra bits used for the atom (pop16 axons only).
     """
 
     spike_dtype = np.dtype(
@@ -255,8 +257,40 @@ class LoihiSpikeInput:
             ("core_id", np.int32),
             ("axon_id", np.int32),
             ("atom", np.int32),
+            ("atom_bits_extra", np.int32),
         ]
     )
+
+    @classmethod
+    def add_spikes_to_generator(cls, t, spikes, basic_spike_generator):
+        methods = {
+            0: getattr(basic_spike_generator, d(b"YWRkU3Bpa2U=")),
+            16: getattr(basic_spike_generator, d(b"YWRkUG9wMTZTcGlrZQ==")),
+            32: getattr(basic_spike_generator, d(b"YWRkUG9wMzJTcGlrZQ==")),
+        }
+        time = d(b"dGltZQ==")
+        chip_id = d(b"Y2hpcElk")
+        core_id = d(b"Y29yZUlk")
+        axon_id = d(b"YXhvbklk")
+        atom = d(b"c3JjQXRvbQ==")
+        atom_bits_extra = d(b"YXRvbUJpdHM=")
+
+        for spike in spikes:
+            axon_type = int(spike["axon_type"])
+            kwargs = {
+                time: t,
+                chip_id: spike["chip_id"],
+                core_id: spike["core_id"],
+                axon_id: spike["axon_id"],
+            }
+            if axon_type == 0:
+                assert spike["atom"] == 0, "Atom must be zero for discrete spikes"
+            else:
+                kwargs[atom] = spike["atom"]
+                if axon_type == 16:
+                    kwargs[atom_bits_extra] = spike["atom_bits_extra"]
+
+            methods[axon_type](**kwargs)
 
     def __init__(self):
         self.axon_map = {}  # maps spike_input idx to axon in self.axons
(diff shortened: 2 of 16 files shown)


arvoelke:
Leaving reminder here: after #260 should be able to change the splitter logic to allow any convolution connections for both precompute True and False.

nengo_loihi/hardware/nxsdk_objects.py
```diff
@@ -260,20 +260,30 @@ def set_axons(self, board, nxsdk_board, spike_input):
...
-                    taxon_idx = int(spike.axon_id)
-                    taxon_id = int(tsyn_ids[taxon_idx])
-                    self.axon_map.setdefault(input_idx, []).append(
-                        np.array(
-                            (-1, axon_type, tchip.id, tcore.id, taxon_id, spike.atom),
-                            dtype=self.spike_dtype,
-                        )
+                if spike is None:
+                    # This should not happen, because input slices happen when connecting
+                    # into the spike input. However, if this arises we just skip it.
```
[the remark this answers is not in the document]
hunse: In `Axon.compartment_map`, we allow a compartment to have an entry of -1 to indicate that that compartment doesn't have any output associated with that `Axon` object. If you call `Axon.map_spikes` with that compartment index, you'll get a `None` spike back. So if we encounter `spike is None` here, it just means that that compartment isn't making use of this `Axon`, so we skip it.

However, when I was testing, I realized that because the way we do our slices if e.g. you've got a sliced pre object off-chip going to a post object on-chip, the slice happens before the `SpikeInput`, so the axon coming from the `SpikeInput` never has compartments with a negative entry in `compartment_map`.

I guess we could test this by manually setting up a `SpikeInput` with a custom `Axon` that does have negative entries for some compartments. It is a feature that the builder could theoretically use, it's just that as it currently happens, it doesn't. So maybe we just need more of a unit test rather than an integration test for this function.

nengo_loihi/hardware/nxsdk_objects.py
```diff
@@ -318,7 +316,7 @@ def set_axons(self, board, nxsdk_board, spike_input):
                             axon.pop_type,
                             tchip.id,
                             tcore.id,
-                            taxon_id,
```
hunse: My reasoning for writing it that way originally was to help make it clear what that parameter is/represents.

nengo_loihi/hardware/nxsdk_objects.py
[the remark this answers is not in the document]
tbekolay: Why do we add the compartment to the map if there are no axons associated with it?

nengo_loihi/hardware/nxsdk_objects.py
tbekolay: It gets changed in #246 anyway which will be squashed in here so it gets changed back.

nengo_loihi/config.py
```diff
@@ -11,6 +11,12 @@ def add_params(network):
       * ``on_chip``: Whether the ensemble should be simulated
         on a Loihi chip. Marking specific ensembles for simulation
         off of a Loihi chip can help with debugging.
+    `nengo.Connection`
+      * ``pop_type``: The axon format when using population spikes, which are only
+        used for convolutional connections. By default, we use ``pop_type`` 32.
+        Setting ``pop_type`` to 16 allows more axons to fit on one chip as long as
+        the ``Convolution`` transform has ``channels_last=True`` and ``n_filters``
+        is a multiple of 4.
```
hunse: We can actually use `pop_type=16` for any type of convolutional connection, it just requires making up to 3 additional copies of the weights to deal with the different offsets. I implement this in dce42a6e.

tbekolay: I just copied this info from the former changelog entry that seemed to indicate that the space savings would come into effect in this situation. Is that not the case? If it is always advantageous to use `pop_type=16` then why don't we just do that and remove all the complexity added to support both types?

hunse: Looking at the tests in `test_conv`, it looks like a lot of them use `n_filters` as a multiple of 4, a holdover from when we did have this restriction. We should probably change some of them to use different values (or test a few values). I'm looking particularly at `test_chip_population_axons` and `test_conv_deepnet`.

tbekolay: That makes sense. Can you answer my question about whether `pop_type=16` is always better to use or are there advantages to `pop_type=32`?

hunse: Looking at dce42a6 again, I think it just implements the copying for cxBase > 256, not for pop16 when `cxBase % 4 != 0`. There's no reason we can't implement the same copying trick to get pop16 to work everywhere, but I don't think I've actually done that anywhere yet.

The reason we don't want to use pop16 always is that the weight copying requires more weight memory. So it's a tradeoff: fewer axons but more weight memory with pop16, or more axons but less memory with pop32. Or you can set up your network in a way that will work with pop16 without any copying (namely `channels_last=True` and `n_filters % 4 == 0` for all layers).

nengo_loihi/hardware/nxsdk_objects.py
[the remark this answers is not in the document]
hunse: Because the map is just a list/array, so it needs an index for each compartment. We could implement it as a dict, but I think I've been making use of the ability to get values for multiple indices simultaneously with an array in some places.

nengo_loihi/hardware/snips/nengo_io.c.template
```diff
@@ -115,10 +117,13 @@ void nengo_io(runState *s) {
...
                axon_id, axon_type, atom);
 #endif
-        if (axon_type == {{ obfs.axon_type_0 }}) {
+        if (axon_type == 0) {
             {{ obfs.do_axon_type_0 }}(s->{{ obfs.step }}, core_id, axon_id);
-        } else if (axon_type == {{ obfs.axon_type_1 }}) {
-            {{ obfs.do_axon_type_1 }}(s->{{ obfs.step }}, core_id, axon_id, atom, 0, 0, 0);
+        } else if (axon_type == 32) {
+            {{ obfs.do_axon_type_32 }}(s->{{ obfs.step }}, core_id, axon_id, atom, 0, 0, 0);
+        } else if (axon_type >= 16) {
```
kinjalpatel27: Is `axon_type > 16` a valid selection?

hunse: Yes. For pop16 axons I'm setting the `axon_type = 16 + extra_atom_bits`.

nengo_loihi/hardware/nxsdk_objects.py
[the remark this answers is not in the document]
tbekolay: I looked into this a fair bit but a lot of the classes in `block.py` do similar things as `Axon` (delayed initialization that is not encapsulated in the class, allowing things like mapping spikes that don't exist, etc) and solving all those issues would take too much time, so I made #270 with the changes I made in case we want to deal with that refactoring later.

nengo_loihi/config.py
tbekolay: I just removed the note about space savings as it seemed too complicated. Anyone wanting to change the `pop_type` will presumably be doing it for their own specific reason.

PR #267, opened 2019-12-06
Add LMU example

tbekolay:
This is literally a copy of the Nengo core example, with some additional text to note that this runs on Loihi, and with some changes to parameters to make it run in a reasonable amount of time with reasonable performance. I also bumped up the learning rate a bit and upped `pes_error_scale` to be able to get decent performance running for only 30 simulated seconds. It still might take a bit too long on the emulator for TravisCI, but we'll see. I haven't tried running it on the board yet, but I'll do that next.

For reference, [here are the plots you get at the end with Nengo core](https://www.nengo.ai/nengo/examples/learning/lmu.html) and below are the plots I got running this on the emulator:

fig1

fig2

fig3

fig4

fig5

I'll post the plots from the chip when I have them.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -22,7 +22,10 @@ Release history
 0.11.0 (unreleased)
 ===================
 
+**Added**
 
+- Added Legendre Memory Unit example.
+  (`#267 <https://github.com/nengo/nengo-loihi/pull/267>`__)
 
 0.10.0 (November 25, 2019)
 ==========================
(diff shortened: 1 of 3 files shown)


tbekolay:
Here are the plots from the chip; worked first time without any changes, which is nice! Performance is about the same as emulator, but faster, which is also nice.

fig1-loihi

fig2-loihi

fig3-loihi

fig4-loihi

fig5-loihi

drasmuss:
Updated plots

image

PR #272, opened 2020-02-28
Support NxSDK 0.9.1

tbekolay:
This is the commit from #266 supporting the next version of NxSDK. Since we're not sure when that next version will be released, I separated it out into a separate PR to deal with later.

### DIFF
--- a/nengo_loihi/tests/test_conv.py
+++ b/nengo_loihi/tests/test_conv.py
@@ -7,6 +7,7 @@ from nengo.exceptions import ValidationError
 from nengo_extras.matplotlib import tile, imshow
 from nengo_extras.vision import Gabor
 import numpy as np
+from packaging.version import parse as parse_version
 import pytest
 import scipy.signal
 
@@ -19,6 +20,7 @@ from nengo_loihi.compat import nengo_transforms
 from nengo_loihi.emulator import EmulatorInterface
 from nengo_loihi.hardware import HardwareInterface
 from nengo_loihi.hardware.allocators import RoundRobin
+from nengo_loihi.hardware.nxsdk_shim import nxsdk_version
 from nengo_loihi.neurons import loihi_rates, LoihiLIF, LoihiSpikingRectifiedLinear
 from nengo_loihi.probe import LoihiProbe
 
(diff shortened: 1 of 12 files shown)


hunse:
All tests pass on NxSDK 0.9.5.rc1.

nengo_loihi/tests/test_conv.py
```diff
@@ -585,10 +583,12 @@ def set_partition(partition=os.environ.get("PARTITION", None)):
...
-        # TODO: This case fails in NxSDK 0.9.0 but will be fixed in the next version.
-        # Remove this check once the next version is released.
-        if pop_type == 32:
-            pytest.skip("Pop32 multichip test requires latest NxSDK")
+        if (
+            pop_type == 32
+            and nxsdk_version is not None
+            and nxsdk_version < parse_version("0.9.1")
+        ):
+            pytest.skip("Pop32 multichip test requires NxSDK >= 0.9.5")
```
drasmuss, line 591: The string says 0.9.5 but the check above is for 0.9.1, which one is correct?

hunse: I'm being a little tricky. First off, there hasn't been a full 0.9.5 release, so the current release is `0.9.5-daily-something`, which is less than `0.9.5`. So I couldn't use that. Also, there was a 0.9.1 release planned, and I think the fix for this was already in that, so I also wanted to allow that. But I'd also be fine with anything else that lets the current daily build work. We could even put it in exactly if we want; it's "0.9.5-daily-20191223" (just make sure to use `parse_nxsdk_version`).

hunse: Alternatively, we could just change the skip message to say 0.9.1. I just didn't want to confuse anyone, since it was never really released.

drasmuss: I suspect there's never going to be a 0.9.1 release at this point. But I'm also kind of reluctant to specify daily build numbers in our tests. What about changing the condition to `nxsdk_version < parse_version("0.9.5.dev0")`? That will let the daily builds work, and any future releases >=0.9.5. Technically someone could try to run the tests with a daily build < 20191223, but I doubt anyone is running these tests other than us, and I don't know why we would ever install an old daily build.

hunse: Sounds good.

hunse:
Fixups look good!

PR #275, opened 2020-03-18
Install NengoDL neuron builders on neuron init

### LINKED ISSUE #248: Loihi neuron type `nengo_dl` builders not used by default
`nengo_loihi.builder.nengo_dl` has the builders for `nengo_dl` for `LoihiLIF` and `LoihiSpikingRectifiedLinear`.

If you make a `nengo_loihi.Simulator` and tensorflow is installed, these builders will get properly initialized (see the `install_dl_builders`) call in `Simulator.__init__`. However, if you just import these neuron types from `nengo_loihi` to do training in `nengo_dl`, before you've created a `nengo_loihi.Simulator`, the builders won't be initialized. 

Even worse, these neuron types are subclasses of `LIF` and `SpikingRectifiedLinear` that `nengo_dl` does know how to build, so it just ends up silently trying to build them but not working.

I think that these builders should be initialized as soon as the neuron types are imported from `nengo_loihi.neurons`. Then, things will just work as people expect. This is what we had originally, but then we changed it for some reason, and I can't remember why. (@tbekolay, you're a co-author with me on that commit 025cdf997. Do you remember why?)

If for some reason we can't *guarantee* that the builders will be initialized, then I think we need some way to throw an error if someone tries to use those neuron types in `nengo_dl` (and that error can point the user to how to initialize them).

hunse:
This insures that the NengoDL neuron builders are added even if
a `nengo_loihi.Simulator` has not yet been created (e.g. if they're
being used for training in NengoDL first).

Fixes #248.

### DIFF
--- a/nengo_loihi/builder/tests/test_nengo_dl.py
+++ b/nengo_loihi/builder/tests/test_nengo_dl.py
@@ -2,10 +2,11 @@ import logging
 
 import pytest
 
+from nengo_loihi.neurons import LoihiLIF
 from nengo_loihi.builder import nengo_dl as builder_nengo_dl
 
 
-def test_installer_with_no_dl(monkeypatch):
+def test_installer_with_no_dl(caplog, monkeypatch):
     """Ensures that the installer works as expected with no Nengo DL.
 
     If Nengo DL is installed, other tests will test the installer as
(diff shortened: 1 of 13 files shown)


nengo_loihi/builder/tests/test_nengo_dl.py
```diff
@@ -42,3 +43,13 @@ class MockNoiseBuilder:
     # Second time warns
     with pytest.warns(UserWarning, match="already has a builder"):
         builder_nengo_dl.NoiseBuilder.register(int)(MockNoiseBuilder)
+
+
+def test_install_on_instantiation():
+    nengo_dl = pytest.importorskip("nengo_dl")
+
+    assert LoihiLIF not in nengo_dl.neuron_builders.SimNeuronsBuilder.TF_NEURON_IMPL
```
hunse: Could there be any weird test order stuff with this, e.g. if another test has already run that's installed the builder?

nengo_loihi/builder/tests/test_nengo_dl.py
```diff
@@ -48,6 +48,11 @@ class MockNoiseBuilder:
 def test_install_on_instantiation():
     nengo_dl = pytest.importorskip("nengo_dl")
 
+    if builder_nengo_dl.install_dl_builders.installed:
+        # undo any installation that happened in other tests
+        del nengo_dl.neuron_builders.SimNeuronsBuilder.TF_NEURON_IMPL[LoihiLIF]
+        builder_nengo_dl.install_dl_builders.installed = False
```
hunse: Ahh, fixed here. Awesome!

hunse:
LGTM!

PR #282, opened 2020-04-08
Add updated CIFAR-10 example

drasmuss:
Moves the example from https://github.com/nengo/nengo-examples/pull/24 to the main nengo-loihi docs.

Also updated all the examples to use hyphens instead of underscores, while I was touching the examples.


setup.py
```diff
@@ -41,11 +41,12 @@ def read(*filenames, **kwargs):
     "matplotlib>=2.0",
     "nbsphinx",
     "nbconvert",
-    "nengo-dl",
+    "nengo-dl>=3.2.0",
     "nengo-extras",
     "nengo_sphinx_theme>=0.7",
     "numpydoc>=0.6",
     "sphinx>=1.8",
+    "tensorflow-probability>=0.9.0",
```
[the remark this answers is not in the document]
drasmuss: Yeah no reason that was just what I had installed and I didn't go back and check earlier versions, I'll change it to 0.8.0.

docs/examples/mnist-convnet.ipynb
```diff
@@ -4,7 +4,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "# Convolutional networks\n",
+    "# MNIST convolutional network\n",
```
hunse, line 7: Should we add a note to this notebook (in the preamble) just saying that it doesn't use block splitting, and to take a look at the CIFAR-10 notebook for an example of that?

drasmuss: I added it a little farther down in the notebook, as there seemed to be a natural spot to talk about it when we discuss the weird parallel architecture we use in the MNIST convnet.

hunse:
Fixups LGTM!

PR #289, opened 2020-05-05
BuildErrors specify an object

hunse:
It is almost always the case that a BuildError is associated with
building a particular object, and specifying the object makes it
easier to track down.


tbekolay (approved):
Rebased and pushed a fixup to change the modified strings to f-strings. It's fine to leave the ones already in the codebase, but ideally we would change them all, so might as well have the newly touched ones use f-strings. Will squash and merge when CI finishes.

PR #290, opened 2020-05-13
Added connection to error output in connection.py

studywolf:
In the process of building some of my models I've encountered errors about the connection build but without specific information about which connection is the cause of the problem. This PR just adds the relevant connection printout to the errors thrown in builder/connection.py.


hunse:
This is already done in #289

PR #297, opened 2020-09-10
Allow 'same' padding for convolution

hunse:
This allows `"same"` padding to be used with `nengo.Convolution`. 

TODO:
- [x] More testing (e.g. in `test_conv_deepnet`)

### DIFF
--- a/.nengobones.yml
+++ b/.nengobones.yml
@@ -201,6 +201,8 @@ setup_cfg:
         no ensembles onchip
       test_simulator.py::test_steps:
         no ensembles onchip
+      test_simulator.py::test_sim_reopen:
+        no ensembles onchip
       test_simulator.py::test_time_absolute:
         no ensembles onchip
       test_simulator.py::test_trange*:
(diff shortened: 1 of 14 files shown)


hunse:
Added an error for the case shown in #298

hunse:
This depends on #296

hunse:
Requires nengo/nengo#1648 for the fix to npconv2d padding for the test, so I think it's easier to just merge this as part of #300.

tbekolay (approved):
Made some fixups so that we always test on both Nengo master and release, and changed the `require_partition` helper to a pytest marker as we were only using `require_partition` to have multichip snips work.

We are getting some random test failures even with these changes, but I've run those tests manually on INRC and they pass when things don't get desynced, so I'm going to merge even if we get some desyncs in this final test run.

tbekolay:
It's still intermittently failing and/or stalling, but I'm at least convinced that it's (mostly) INRC environment issues, so I'm going to merge regardless. We can clean up any issues when all the v1.1.0 PRs are in.

PR #301, opened 2020-10-28
Split blocks speedups

hunse:
Speed improvements for `split_blocks.py`.

TODO:
- [ ] I've just commented out two costly assertions. They were mostly there for debugging during during development. We could get rid of them completely, or have an option to turn them on (e.g. when running all the unit tests), or just test that they are true in a specific unit test.


hunse:
Closed in favour of #309.

PR #304, opened 2020-12-03
Add Nengo version compatibility to changelog

hunse:

PR #305, opened 2020-12-22
Limit tensorflow-probability to <0.12

drasmuss:
For compatibility with the tensorflow<2.4 requirement added in https://github.com/nengo/nengo-loihi/pull/304.

### DIFF
--- a/.nengobones.yml
+++ b/.nengobones.yml
@@ -541,6 +541,7 @@ ci_scripts:
       - $NENGO_DL_VERSION
       - cython
       - tensorflow<2.4
+      - tensorflow-probability<0.12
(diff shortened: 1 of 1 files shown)


hunse (approved):

hunse:
Shouldn't the new pip dependency solver avoid problems like this?

drasmuss:
Yeah I would have thought so, but I think it's not working because we install the dependencies in a couple of stages. In particular we install `tensorflow` and `tensorflow-probability` at different points, so I think it is not picking up on the fact that we have an incompatible version of `tensorflow` installed when it gets to the `tensorflow-probability` stage.

PR #308, opened 2021-01-20
Release candidate 1.0.0

hunse:
I tested some older NxSDK versions. We were failing on 0.8.7 because some of the SNIP communication stuff doesn't work with that, so I've just dropped support for 0.8.7. Everything works on 0.9.0. (Except `test_conv_deepnet`, but that's an expected failure, since there's some population axon stuff that's been fixed in a more recent NxSDK. We actually used to skip that test on older NxSDK, but that got removed at some point. We could re-add an xfail or something, but it didn't seem worth it since we only ever run these tests on older NxSDK manually on a release.)

I also found some of the DVS stuff didn't work with older python/NxSDK, so I've fixed that. Also fixed some inconsistent naming of the `format` argument for DVS file format (I know `format` is a built-in, but I've never seen us use it as a stand-alone function, so I figured it's fine to overwrite it).

### DIFF
--- a/nengo_loihi/dvs.py
+++ b/nengo_loihi/dvs.py
@@ -14,7 +14,7 @@ class DVSFileChipProcess(ChipProcess):
     ----------
     file_path : string
         The path of the file to read from. Can be a ``.aedat`` or ``.events`` file.
-    kind : "aedat" or "events" or None, optional
+    file_fmt : "aedat" or "events" or None, optional
         The format of the file. If ``None`` (default), this will be detected from the
         file extension.
     t_start : float, optional
@@ -58,7 +58,7 @@ class DVSFileChipProcess(ChipProcess):
     def __init__(
         self,
         file_path,
-        format=None,
+        file_fmt=None,
         t_start=0,
         rel_time=None,
         pool=(1, 1),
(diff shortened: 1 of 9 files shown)


nengo_loihi/dvs.py
```diff
@@ -14,7 +14,7 @@ class DVSFileChipProcess(ChipProcess):
     ----------
     file_path : string
         The path of the file to read from. Can be a ``.aedat`` or ``.events`` file.
-    kind : "aedat" or "events" or None, optional
+    format : "aedat" or "events" or None, optional
```
tbekolay, line 17: It would be good to use one set of terminology everywhere; is there an external spec or something that we are trying to match? If not, I would want to avoid the use of `format` since it is a Python builtin. I'm not a huge fan of `kind` either, could we do `file_fmt`? `fmt` is typically use for a string formatting template.

tbekolay (changes requested):
Would like to avoid shadowing the Python builtin `format` if we can avoid it; will leave it to @hunse to change to a name that he's OK with.

nengo_loihi/dvs.py
hunse: `format` is a built-in, but as I said above, we never use it. There are also examples in Nengo/Numpy/TensorFlow where built-ins are overridden locally.

But I'm also fine with `file_fmt`, so I'll change it to that.

Issue #313, opened 2021-03-12
Cifar10 Example produces attribute error in TF compile call


xchoo:
To replicate this error: 
- Run the Cifar10 example 

The error:
- The error is visible in the [online docs](https://www.nengo.ai/nengo-loihi/examples/cifar10-convnet.html#Train-network-using-NengoDL), cell 8:
```
Could not compute ANN values on this machine: in user code:

    /home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/tensorflow/python/keras/engine/training.py:1224 test_function  *
        return step_function(self, iterator)
    /home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/tensorflow/python/keras/engine/training.py:1215 step_function  **
        outputs = model.distribute_strategy.run(run_step, args=(data,))
    /home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/tensorflow/python/distribute/distribute_lib.py:1211 run
        return self._extended.call_for_each_replica(fn, args=args, kwargs=kwargs)
    /home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/tensorflow/python/distribute/distribute_lib.py:2585 call_for_each_replica
        return self._call_for_each_replica(fn, args, kwargs)
    /home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/tensorflow/python/distribute/distribute_lib.py:2945 _call_for_each_replica
        return fn(*args, **kwargs)
    /home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/tensorflow/python/keras/engine/training.py:1208 run_step  **
        outputs = model.test_step(data)
    /home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/tensorflow/python/keras/engine/training.py:1177 test_step
        y, y_pred, sample_weight, regularization_losses=self.losses)
    /home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/tensorflow/python/keras/engine/compile_utils.py:187 __call__
        self.build(y_pred)
    /home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/tensorflow/python/keras/engine/compile_utils.py:140 build
        self._losses = nest.map_structure(self._get_loss_object, self._losses)
    /home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/tensorflow/python/util/nest.py:635 map_structure
        structure[0], [func(*x) for x in entries],
[...]

PR #315, opened 2021-04-20
Fixed functools.partial fix in Cifar10 notebook

xchoo:
Replaced functools.partial call with proper function definitions. Addresses issue #313.


hunse:
Why?

xchoo:
Updated description with issue number.

hunse:
Also now setting the learning phase using the NengoDL method that works with newer TensorFlow.

tbekolay (approved):
LGTM, runs locally without (unexpected) warnings. I also added the params.npz file to `.gitignore`. Merging when CI is happy.

PR #318, opened 2021-11-17
Support NxSDK 1.0.0 and drop Python 3.5 support

hunse:
This allows us to use NengoBones `master` again.

I have not removed `OrderedDict`, switched over to f-strings, or switched to `pathlib`. All of these seem like a decent amount of work with little/no benefit, so not worth doing at this time.


tbekolay (approved):
This LGTM! I added two small fixups, one fixing some Pylint issues (without disabling the check) and one modifying the installation instructions. When CI finishes I'll squash and merge.

docs/installation.rst
```diff
@@ -30,7 +30,7 @@ for running NengoLoihi,
 as well as for running models
 using the NxSDK directly.
 
-Note, you *must* use Python 3.5.2 when working with NxSDK.
+Note, you *must* use Python 3.8 when working with NxSDK.
```
hunse, line 33: I'm not sure if this is actually a requirement. I'm using 3.7 and it seems to work fine. I think our hardware tests actually use 3.6.

tbekolay: Oh nice, that simplifies things. I know they don't have .pyc file anymore so that makes sense :+1:

PR #320, opened 2021-11-25
Deobfuscate

hunse:
(based off #289)

Yay! We are no longer obfuscated!

I think this can probably all be squashed into one commit, I just wanted to keep it separate for testing (and maybe it's helpful for review, too?)

Most of the deobfuscation was performed by our automated script, but there were a few places where I had to step in with some manual changes (the biggest one being the snips).

I also went through the obfuscation commit (bcebef3e5bf59cf05fdf487b7f9dfd9a70a60688), and found a few places where we were using a more "explicit" representation of a number (e.g. `2 ** 12 - 1` rather than 4095), and changed those back.


tbekolay (approved):
Looks great, excited to have this in once CI finishes!

Issue #323, opened 2021-12-03
Possible memory leak for multiple run_steps()


nmnaughton:
Hi, following up on #311. I tried using the new code on the master branch and was still having issues. Playing around with the test script from  #312 I think the issue is that when `sim.run_steps()` is called many times in a row a memory leak occurs. I am defining a memory leak here to be the resident memory associated with the process as seen in `top`.  In the below code, when the for loop is called many times in a row, after each `.run_steps()`, a section of memory is taken up proportional to the number of time steps just called. This stacks up throughout the run and can result in a lot of memory. If `.run_steps()` is only called a few times, then this memory gets released when `gc.collect()` is called. If a lot of `.run_steps()` is called, then this memory does not get released correctly and stays reserved (Case 1 vs Case 2 below).

So I think that the lack of releasing the memory is likely a bug, but is it also possible to not have that memory reserved at all? in my use case, I am trying to run a reservoir for a long time, and even if the memory is correctly released at the end of the simulation, currently it takes up ~20 Gb of memory due to the length of the simulation and I can not scale the network any further. I have no probes in my simulation and do not to need any past information about the reservoir, but currently the memory requirement is proportional to how many time steps I have taken, which seems like unwanted behavior. 

Thanks! 

```python
import gc
import weakref

import nengo
import numpy as np

import nengo_loihi


class NewClass:
    def __init__(self):
        self.input_size = 10
        self.n_neurons = 1024
        self.initialize_nengo()

    def initialize_nengo(self):
        network = nengo.Network()
        with network:

            def input_func(t):
                return np.ones(self.input_size)

            def output_func(t, x):
                self.output = x

            input_layer = nengo.Node(
                output=input_func, size_in=0, size_out=self.input_size
            )

            ensemble = nengo.Ensemble(
                n_neurons=self.n_neurons, 
                dimensions=1, 
            )
            
            output_layer = nengo.Node(
                output=output_func, size_in=self.n_neurons, size_out=0
            )

            conn_in = nengo.Connection(
                input_layer,
                ensemble.neurons,
                transform=np.ones((self.n_neurons, self.input_size)),
            )
            conn_out = nengo.Connection(ensemble.neurons, output_layer, synapse = None)

        self.network = network

    def run(self, case, num_resets):
        for i in range(num_resets):
            with nengo_loihi.Simulator(self.network, precompute=True) as sim:
                if case == 1:
                    """ Memory increases throughout the run and at the end the total memory that has been
[...]

PR #324, opened 2021-12-06
Clear spikes

hunse:
Clears intermediary values as we consume them, freeing up memory (particularly when `run_steps` is called more than once). Addresses #323.

### DIFF
--- a/LICENSE.rst
+++ b/LICENSE.rst
@@ -4,7 +4,7 @@
 NengoLoihi license
 ******************
 
-Copyright (c) 2018-2021 Applied Brain Research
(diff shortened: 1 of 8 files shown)


hunse:
I added the `permanent` flag to ensure that some spikes (i.e. those added during build) can be kept permanently in the `SpikeInput`. This isn't strictly necessary right now, but will be if/when we allow simulator resetting, so figured this fix makes the most sense.

tbekolay (approved):
LGTM, not sure why we were keeping the queued items around in the first place!

PR #325, opened 2022-01-04
Remove examples and NengoDL builders

tbekolay:
These have been moved to NengoExamples and NengoExtras.


tbekolay:
This PR is now ready for review. It won't pass CI until https://github.com/nengo/nengo-bones/pull/160 and https://github.com/nengo/nengo-examples/pull/28 are done and merged, but nothing in this PR will change if those PRs don't change significantly.

nengo_loihi/tests/test_basic.py
```diff
@@ -0,0 +1,154 @@
...
+    plt.plot(t, a)
+    plt.plot(t, b)
+
+    assert anp in sim.data
+    assert avp in sim.data
+    assert allclose(a, 0.0, atol=0.03)
+
+    assert bup in sim.data
+    assert bvp in sim.data
+    firing_rate = np.sum(sim.data[bnp], axis=0) * sim.dt / sim.n_steps
```
hunse: Technically, the firing rate in Hz is without multiplying by `sim.dt` (which can be written more simply as just `sim.data[bnp].mean(axis=0)`). Then you'll get an average firing rate that's in a more normal range (30-50 Hz).

tbekolay: Yeah I made the name before multiplying by `sim.dt` (was doing that below before), then changed and didn't update the name. Feel free to change :+1:

tbekolay:
Updated this with the latest NengoBones. Your commits LGTM @hunse :+1:

hunse (approved):
LGTM.

The hardware tests errored, and I would like to see them pass because I made that change to the DecodeNeurons scale factor. If they error again, though, I might try to just run them manually.

setup.py
```diff
@@ -32,28 +32,20 @@ def read(*filenames, **kwargs):
 install_req = [
     "jinja2",
     "nengo>=3.1.0",
+    "nengo-extras>=0.5",
```
hunse, line 35: Thinking more about this; do we need to do this? It's only required to use the NengoDL builders, and if someone tries to use those neurons with NengoDL without NengoExtras, they'll get an error.

My biggest concern with this is it's one more dependency that we have to keep up to date both with this and with core Nengo. That said, we *should* be keeping it up to date, so as long as we stay on top of it, I'm OK.

tbekolay: We don't need to do it, but not doing makes it super fragile and easy to introduce cyclic dependencies. If you want to remove it and go through all that again you can try, but I would really like to get this release done, and it might require changes to NengoExtras that will need to get released again. NengoExtras itself is not hard to install once Nengo is installed so I'd really prefer to leave it, as it was a big time sink to get it working without cyclic dependencies.

hunse: I don't quite follow why it would be so fragile. I thought it's just a matter of moving it from `install_req` to `tests_req` in this file (since we do use it for tests). The import in the main code (in `neurons.py`) is already guarded against NengoExtras not being installed.

If it would be hard to change it, though, I'm fine to keep it here. I agree, we want to get the release out.

hunse: Was easy to change, so I've moved it to "optional" and "tests" :rocket:

